PR #2, opened 2026-09-01
blip-check: probe every Contacts source, not just glob()[0]

jethrojones:
## Problem

`check_contacts()` in `bridge/mac/blip-check` opens **one** `.abcddb` — `glob.glob(AB_GLOB)[0]`, and `glob` order is filesystem order, not sorted — and reports the entire contacts check as failed if that single DB does not open.

Macs accumulate disabled / stale CardDAV account sources whose `AddressBook-v22.abcddb` never opens (`unable to open database file`) no matter what TCC grants exist. The `contacts` tool already handles this — it `warn: skipping …` and moves on. But when one of those dead sources happens to sort first, `blip-setup` prints:

```
❌  contacts   unable to open database file
    fix: Same Full Disk Access grant as chat.db …
```

The suggested fix does nothing (FDA is already granted — `chatdb` passes), and `contacts list` / avatars over the bridge work fine. On a real machine here: 7 of 16 source DBs open, `contacts list` returns real contacts, but `blip-check` reported `contacts ❌` because a dead source sorted first.

## Fix

- Iterate **all** sources — `sorted(glob(...))` plus the root `AddressBook-v22.abcddb`, matching `contacts`' own `source_dbs()`.
- Sum records across the DBs that open; fail only when **nothing** opens (the real missing-FDA signal).
- Detail line now reads e.g. `13950 contacts readable (7/16 sources, 9 stale source(s) skipped)` so stale sources are visible but not alarming.

Deterministic now (sorted), so no more order-dependent flapping.

## Test

```
$ python3 -m py_compile bridge/mac/blip-check   # CI byte-compile: OK
```

Local harness with a dead source sorted ahead of a good one:

```
first source (old code would test only this): AAA/AddressBook-v22.abcddb
NEW check_contacts -> (True, '42 contacts readable (1/2 sources, 1 stale source(s) skipped)')
all-dead          -> (False, 'file is not a database')
```

Also verified against a live bridge: `blip-check` now reports contacts ✅ where it previously reported ❌.

## Note

`bridge/mac/tcc-check`'s `contacts_db()` has a related weakness (loops all sources but raises on the first that fails). Left out of this PR to keep it focused on the `blip-setup` path; happy to follow up.

### DIFF
--- a/bridge/mac/blip-check
+++ b/bridge/mac/blip-check
@@ -20,7 +20,8 @@ import subprocess
 import sys
 
 CHAT_DB = os.path.expanduser("~/Library/Messages/chat.db")
-AB_GLOB = os.path.expanduser("~/Library/Application Support/AddressBook/Sources/*/AddressBook-v22.abcddb")
+AB_DIR = os.path.expanduser("~/Library/Application Support/AddressBook")
+AB_GLOB = os.path.join(AB_DIR, "Sources", "*", "AddressBook-v22.abcddb")
 
 FIX = {
     "chatdb": "System Settings → Privacy & Security → Full Disk Access → add "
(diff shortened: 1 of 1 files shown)


nixfred (approved):
Verified on a live bridge: the patched check reports `13484 contacts readable (4/4 sources)` where the old code was reading a single, unordered `glob()[0]`. The iteration mirrors `contacts.source_dbs()` exactly (sorted sources + root db), and failing only when *nothing* opens is the right FDA signal. Clean, focused, well explained — thank you. Merging as-is; credited in CHANGELOG and README.

The `tcc-check` follow-up you noted is yours if you want it — same shape, same rationale.

PR #4, opened 2026-09-02
Send SMS/RCS DMs on the matching service

lukejmorrison:
## Why

2.2.0 already sends SMS/RCS when `thread.service` is already SMS/RCS. A **new** chat to an Android number still starts as iMessage. AppleScript does not fall back the way Messages.app GUI does, so the send fails with **error 22** (not registered for iMessage). The failed bubble keeps `service: iMessage`, so the next Enter retries iMessage.

This is the leftover from the ROADMAP item 2.2.0 closed: last-inbound (and failed-iMessage-to-phone → SMS).

Useful if Omarchy (`omacom/omarchy`) vendors Blip: first-hour SMS to a non-Apple phone currently looks broken.

## What

Rebased onto **2.2.2**.

- Thread `service` is last **inbound**.
- If there is no inbound and the only rows are **failed iMessage to a phone number**, next send is **SMS**.
- Email handles stay iMessage.
- Compose placeholder follows (`SMS` while the failed bubble is on screen).
- Groups still send `--chat-id` only.

New chats with no history still try iMessage first (same as Messages.app). After error 22, the next send is SMS. SMS still needs Text Message Forwarding on the paired iPhone.

## Tests

`bun test` — 204 pass. Logic in TypeScript (`sendServiceForMessages`); QML only binds. Injected runners only; no live send to a real contact.

## Invariant

DMs send `--to <chat>` plus `--service SMS|RCS` when last inbound (or a failed iMessage to a phone) says so. Message bodies stay on stdin.

### DIFF
--- a/BlipView.qml
+++ b/BlipView.qml
@@ -569,6 +569,20 @@ FocusScope {
     }))
   }
 
+  /** SMS/RCS, or SMS after a failed iMessage to a phone (error 22). */
+  function sendService() {
+    var chat = String(root.active && root.active.chat ? root.active.chat : "")
+    var phone = /^\+?[0-9]{5,}$/.test(chat)
+    if (phone) {
+      for (var i = 0; i < root.bubbles.length; i++) {
+        if (root.bubbles[i].from_me === true && root.bubbles[i].failed === true) return "SMS"
+      }
+    }
+    var s = String(root.active && root.active.service ? root.active.service : "")
+    if (s === "SMS" || s === "RCS") return s
+    return "iMessage"
+  }
+
   function send() {
     var text = composeField.text
     if (!root.inThread) return
@@ -602,7 +616,7 @@ FocusScope {
       sendDraftPath = draftPath
       // caption on stdin — never in this process's argv (audit #4, war room #1/#13)
       fileSendProc.command = ["bun", root.sendFileScript, sendChat, draftPath, "--caption-stdin"]
-        .concat(/^(SMS|RCS)$/i.test(String(root.active.service || "")) ? ["--service", String(root.active.service).toUpperCase()] : [])
+        .concat(/^(SMS|RCS)$/i.test(root.sendService()) ? ["--service", root.sendService()] : [])
       fileSendProc.stdinEnabled = true
       fileSendProc.running = true
       fileSendProc.write(trimmed !== "" ? text : "")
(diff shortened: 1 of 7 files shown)


nixfred (changes requested):
Thanks Luke — and sorry about the timing: 2.2.0 (3c4e65d) landed the `--service` pass-through about an hour after you opened this, which is why the diff now conflicts in BlipView.qml and send-file.ts.

The part main does NOT have is the good part: `sendServiceForMessages` picking the service from the last **inbound**, so a failed iMessage to an SMS contact can't flip the thread back to iMessage. Main still uses the last bubble's service, so that stuck-thread case is real on 2.2.0.

Would you rebase onto main and slim this down to that rule plus its tests — `normalizeSendService` / `sendServiceForMessages` in collector.ts, the one-line change in `buildThreads`, and the six tests? The QML and send-file plumbing is already there.

The Copilot note about captions in argv is pre-existing (BlipView always uses `--caption-stdin`; the argv path is the CLI fallback) — not on you.

One thing for the record: I can't find any thread in omacom/omarchy about Blip as core bar support. If you have a link, I'd like to see it.

nixfred (changes requested):
Much better — rebased clean on 2.2.2, 204 pass, the `--to` line is intact, and the placeholder fix is welcome (the 99123 SMS thread was showing "iMessage"). Two things, both in the same spot:

**1. Drop the bubble scan in QML `sendService()`.** It returns SMS for a phone chat as soon as *any* failed outbound sits in the loaded window, before it looks at the service — so a contact who is on iMessage, with one old "Not Delivered" bubble in the last 80 rows, gets every send forced to SMS. That contradicts your own headline rule (last inbound wins). It also isn't needed: a failed iMessage row lands in chat.db with `error` set, the poll picks it up within a second, `last_ts` changes, and `onThreadsChanged` swaps `active` for the fresh thread object, so `active.service` already carries the collector's answer. Bind the placeholder and the send to `root.active.service` (normalized) and keep the logic in TypeScript, per the house rule. The `ui.test.ts` line asserting the scan goes with it.

**2. Put the case you were reaching for into `sendServiceForMessages` instead.** The real gap is "contact left iMessage": last inbound is an old iMessage, the newest row is our failed iMessage (error 22) to a phone number — today the rule returns iMessage forever. Check that first: *if the newest row is a failed `from_me` iMessage and the chat is a phone → SMS*; then last inbound; then the rest as you have it. One test for it ("inbound iMessage, then failed iMessage 22 to a phone → SMS"); your existing "SMS inbound then failed iMessage → SMS" still holds under that order.

Push that and I'll merge.

nixfred:
Landed in 1d0476a, with you as co-author — thank you, and sorry it took a detour.

Your rule was right and I said so twice; what I kept pushing back on was where it lived. The QML `sendService()` returned SMS for **any** phone chat with **any** failed outbound in the loaded window, so a contact who is on iMessage with one old "Not Delivered" bubble from months ago would have had every send forced to SMS. That contradicted your own headline rule, and it was invisible from the diff unless you thought about a stale bubble.

So the shipped version is your logic in TypeScript, with the ordering I asked for:

1. our newest row is a failed iMessage to a phone → SMS (your error-22 case)
2. otherwise the last **inbound** service — your original insight, and the one that matters most
3. no inbound → the last outbound that succeeded
4. nothing → iMessage

Plus the guard you did not have: an email handle can only be iMessage, so a failed iMessage there does not become SMS. Groups keep the raw value since they send by chat-id.

Measured on my own Mac after the change: 136 DM threads now send SMS, 16 RCS, 105 iMessage, and all 42 group threads unchanged.

I am closing this PR rather than leaving it open, because your branch is 37 commits behind — today the repo took 8 PRs from 6 people plus a release — and the bug was worth fixing now rather than waiting on a rebase. That is not a judgement on the work, it is the opposite: the work was good enough to carry forward without you.

The compose-placeholder change from your second round I did not take, only because the same information now comes from `thread.service` and I wanted one source. If you want to send that as its own small PR against `main`, I will take it. And thank you for the original ROADMAP nudge that started this — the `--service` pass-through in 2.2.0 was yours too.

PR #5, opened 2026-09-02
toasts: notify new inbound as system notifications by default

nova-centauri:
## What

New inbound iMessages fire a **system notification** (Omarchy's notification daemon via `notify-send`) instead of only lighting the tiny bar badge.

`~/.config/blip/allowlist.json` is now a toast *policy*, not a required gate:

- missing file → notify everyone (new default)
- `{ "mode": "off" }` → badge only
- `{ "mode": "allow", "allow": ["+1…"] }` → those senders
- legacy `{ "allow": ["+1…"] }` still filters; legacy `{ "allow": [] }` still means off

Also hardened: phone/email matching, skip the open conversation and the self-thread, cap a catch-up at 20, attachment-only rows say "New message", group toasts read `Alice · Family`, corrupt policy file fails closed.

## Why

The bar badge is easy to miss. Toasts were allowlist-gated, and an empty/missing file meant silence, so most installs never interrupted. Clicking the toast still opens that conversation with compose focused.

If you liked the old quiet default:

```json
{ "mode": "off" }
```

## How it was verified

- [x] `bun test` is green (CI will check) — **206 pass**, 0 fail
  - this change does not send; collector logic is unit-tested without `imsg-send`
  - `notify-send --app-name=Blip` was exercised locally against Omarchy's daemon
  - notifications are the system daemon's card, not a Blip layout change

Invariants touched:
- **Pass `--` before message text** on `notify-send` (`notifySendArgv` + `ui.test.ts` contract)
- **Two read marks** — toasts still keyed off `watermark`, not `readMark`
- **No message content in state.json** — still opaque SHA-256 toast keys only
- **Toasts are a policy** — new invariant documenting mode=all/allow/off, open-chat skip, batch cap, `toastPreview`

### DIFF
--- a/BarWidget.qml
+++ b/BarWidget.qml
@@ -11,9 +11,10 @@ import qs.Ui
 // widget owns the single poller; the panel reads its state rather than running
 // its own collector.
 //
-// Badge counts EVERY unread. Toasts fire only for allowlisted handles
-// (~/.config/blip/allowlist.json) — chat.db is mostly bank alerts and 2FA codes,
-// and none of that deserves an interruption.
+// Badge counts EVERY unread. Toasts fire for every new inbound as a system
+// notification. An optional ~/.config/blip/allowlist.json restricts that to
+// listed handles — chat.db is mostly bank alerts and 2FA codes, so the
+// allowlist is how you quiet the noise without losing the badge.
 //
 // Left-click = panel · double-click = the app window · middle-click = refresh · right-click = mark all read.
 BarWidget {
@@ -417,12 +418,14 @@ BarWidget {
   }
 
   // ------------------------------------------------------------ toasts
-  // One notify-send per allowlisted inbound message. The collector has already
-  // applied the allowlist and the dedupe ring, so anything arriving here is
-  // meant to interrupt. Reuses the no-focus-steal notification behaviour.
+  // One notify-send per Toast the collector emitted. Policy (all / allow /
+  // off), open-chat skip, self-thread skip, preview sanitizing, and the
+  // batch cap all live in collector.ts — this just renders. Reuses the
+  // no-focus-steal notification behaviour.
   // With --wait + actions, clicking the toast (default) or its Reply button
   // opens the panel ON that conversation, compose focused — the daemon
   // advertises the "actions" capability (verified via GetCapabilities).
+  // Flag set must match collector.notifySendArgv (ui.test.ts).
   Process {
     id: notifyProc
     stdout: StdioCollector {
@@ -458,7 +461,9 @@ BarWidget {
       "notify-send",
       "--app-name=Blip",
       "--icon=mail-message-new",
-      "--expire-time=8000",
+      "--urgency=normal",
+      "--category=im.received",
+      "--expire-time=15000",
       // --wait blocks until the toast closes and prints the chosen action.
       // Only `default` (clicking the card): the Omarchy notification daemon
       // renders no action BUTTONS and only ever invokes default — an
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,37 @@
 # Changelog
 
+## 2.0.4 — 2026-09-01
+
+**Changed**
+- New inbound messages fire a **system notification** by default (Omarchy's
+  notification daemon via `notify-send`). The tiny bar badge is no longer
+  the only alert. Click the toast to open that conversation.
+- `~/.config/blip/allowlist.json` is now a toast *policy*, not a required
+  gate. Missing file = notify everyone. `{ "mode": "off" }` is silence.
+  A populated `{ "allow": ["+1…"] }` still filters to those senders.
+
+**Upgrade note.** If you liked the old quiet default (badge only, no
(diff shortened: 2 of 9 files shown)


nixfred (changes requested):
Thanks Nova — this is careful work and the tests are exactly the kind Blip wants. Merged onto 2.2.0 locally it is green (220 pass); the conflicts are docs and manifest only.

Two changes before it can land:

1. **Keep the quiet default.** A missing `allowlist.json` must stay badge-only. That is a deliberate rule: chat.db is mostly bank alerts and 2FA codes, and a fresh install must not fire 20 cards on its first catch-up. Make `{ "mode": "all" }` the opt-in for "everyone" and keep everything else — mode all/allow/off, the open-conversation and self-thread skips, `TOAST_BATCH_CAP`, fail-closed on a corrupt file, `foldHandle`, `notifySendArgv` locked by ui.test. README / PRIVACY / CLAUDE.md then describe the opt-in instead of a new default. Please also keep `--expire-time` at 8000; 15 s cards are long.
2. **Drop the manifest bump.** 2.0.3→2.0.4 is against a main that is at 2.2.0; versions are cut at release time here.

Rebase onto main — 2.2.0 changed the toast path (leader-only toasts on multi-monitor, failure-toast keys) — and I'll merge.

nixfred:
Status, and an honest one: this is still open because it is the only PR here I would not land on your behalf.

The hardening in it I want, essentially all of it — `mode` all/allow/off, the open-conversation and self-thread skips, `TOAST_BATCH_CAP` so a catch-up after sleep cannot dump 150 cards, fail-closed parsing, `foldHandle` so `+15550100011` and `(555) 010-0011` are the same person, and `notifySendArgv` locked to the QML by a test. Merged onto main a day ago it was green at 220 tests. None of it is in `main` today; I checked before writing this.

What I will not do is decide the default for you. Flipping "missing allowlist" from badge-only to notify-everyone is a product call, it is the heart of your PR, and rewriting it myself would be me disagreeing with your design while wearing your name. So it waits for you.

Two things have changed underneath it since:

- `main` moved 58 commits, including a release (2.3.0). The toast path itself changed: toasts are now leader-only on multi-monitor, failure-toast keys carry a `fail:` prefix through the dedupe ring, and `normalizeToastKey` gained a `link:` namespace. Your `selectToasts` signature change will need to meet those.
- Announcement rows (`item_type != 0`) are filtered at the bridge now, which removes a class of empty toast your batch cap was partly protecting against. There were 811 of those on my Mac.

If you would rather not rebase 500 lines through that, say so and I will do the rebase myself and merge it under your name with the quiet default kept — I just want you to choose, not find it done. No rush and no hard feelings either way; the work is good.

nixfred:
Checking in, and the short version is: still wanted, still open, no deadline from me.

Main has moved a fair way — seven PRs landed over the last two days, including a compose/send rewrite — so this branch now conflicts. That is on the passage of time, not on you.

Nothing about my position has changed. I want the hardening in this PR: `mode` all/allow/off, the open-conversation and self-thread skips, `TOAST_BATCH_CAP` so a catch-up after sleep cannot dump 150 cards, fail-closed parsing, and `foldHandle` so `+15550100011` and `(555) 010-0011` are one person. That is careful work and Blip has nothing like it.

The one blocker is still the single line from my review: **a missing `allowlist.json` must stay badge-only.** `chat.db` is mostly bank alerts and 2FA codes, and a fresh install must not fire twenty cards on its first catch-up. Make `{"mode": "all"}` the explicit opt-in for everyone and keep the rest exactly as you have it.

If you want to pick it back up, rebase onto current main and flip that default and I will land it. If you would rather not, say so and I will lift the hardening into a separate change with credit to you — I would rather have your name on it, but I would rather have the code than neither.

nixfred:
Thank you for this, and sorry it sat — it deserved an answer months sooner than 131 commits later.

Closing it, but I want to be precise about why, because the work wasn't wrong.

**The mechanism landed.** `main` now fires `notify-send` per inbound message with a dedupe ring, a 20-toast cap on a hung `notify-send`, click-and-Reply actions that open the panel on that conversation with compose focused, and no focus steal. Several of those are things you identified here independently — the cap in particular.

**The default flip is the one thing I'm deliberately not taking.** `BarWidget.qml:15` says why: the badge counts every unread, but toasts stay allowlist-gated because `chat.db` is mostly bank alerts and 2FA codes. Notifying everyone by default makes the common install noisy enough that people turn toasts off entirely, which costs them the messages that mattered. That's a product call, not a judgement on the patch.

**One idea in here I do still want.** Making `allowlist.json` a *policy* — `{"mode": "off" | "allow"}` with the legacy bare-`allow` shape still honoured — is better than a bare gate, and it's orthogonal to what the default is. If you'd like to send that as a small standalone PR against current `main`, keeping the quiet default, I'll take it.

The branch is 131 commits behind with 8 conflicting files, so rebasing this one wholesale isn't worth your time. Thanks again for the care in the write-up — the legacy-shape compatibility table is the part that made the policy idea obviously right.

nixfred:

PR #6, opened 2026-09-02
fix: bar double-click + Android SMS/RCS

washburnello:
Fixes two user-facing bugs found after Tailscale setup

**Bar double-click opening app**
- `BarWidget.qml` handled double-click via a Timer on `onClicked` (`dblClick 320ms`). Omarchy's `Bar.qml:1477` `CenterGestureArea.onDoubleClicked` toggles `transparent` on empty-space double-click. `WidgetButton` only emitted `pressed` on `onClicked`, never `onDoubleClicked`, and `ModuleSlot.modulePointer` (which sits atop the widget) only handled `onClicked` via `pressModuleClickTarget`. Double-clicks fell through to the bar background → bar flashed transparent instead of opening the app.
- Fix: `Ui/WidgetButton.qml` adds `signal doublePressed` and `onDoubleClicked: {accept; doublePressed}`, `Bar.qml` `modulePointer.onDoubleClicked` forwards to `target.doublePressed` and always `accept`s to block `CenterGestureArea`, `BarWidget.qml` adds `suppressNextLeftClick` to swallow the pending second `leftClick` after the native double.

**Android SMS/RCS not sending**
- `BlipView.qml:611` always called `imsg-send --to <handle> --yes` defaulting to `iMessage` (`imsg-send:434`). Threads from `collector.ts:386` carry `service` (`SMS`/`RCS`/`iMessage`), `chats` shows duplicate ids per service (e.g. `+16789671419` has `RCS 1240` and `SMS 8102`). Android contacts on `SMS`/`RCS` failed silently (`error` → `selectFailures`).
- Fix: `BlipView.qml` honors `active.service` → `--service SMS/RCS/iMessage` for both text and file sends; `send-file.ts` adds `--service` param and threads it to `imsg-send`.

**Also:** `BlipView.qml:288` `openLink` jq filter fix is already upstream (`$2` argv) - included.

Test: `bun test --cwd . ./collector.test.ts` 90 pass, `thread.test.ts` 54 pass, `tier2.test.ts` 36 pass, `ui.test.ts` 6 pass; `qs ipc call nixfred.blip app` → `org.quickshell|Blip` mapped.

System patches (`Ui/WidgetButton.qml`, `Bar.qml:modulePointer`) belong in `basecamp/omarchy` - this PR is the Blip side. Happy to split if preferred.

Closes: bar double-click transparency, Android send via correct service.

### DIFF
--- a/BarWidget.qml
+++ b/BarWidget.qml
@@ -138,7 +138,12 @@ BarWidget {
   // Single click = panel (opens immediately — no double-click lag); a second
   // click within the double-click window = close the panel and open the APP.
   Timer { id: dblClick; interval: 320 }
+  property bool suppressNextLeftClick: false
   function leftClick() {
+    if (suppressNextLeftClick) {
+      suppressNextLeftClick = false
+      return
+    }
     if (dblClick.running) {
       dblClick.stop()
       root.close()
@@ -568,6 +573,14 @@ BarWidget {
       if (code === Qt.MiddleButton) root.refresh(false, false)
       else if (code === Qt.RightButton) root.markAllRead()
       else root.leftClick()
+    }
+    onDoublePressed: function(code) {
+      if (code === Qt.LeftButton) {
+        suppressNextLeftClick = true
+        dblClick.stop()
+        root.close()
+        root.showApp()
+      }
     }
 
     Row {
--- a/BlipView.qml
+++ b/BlipView.qml
@@ -598,7 +598,9 @@ FocusScope {
       // send-file.ts owns target resolution (group guid or DM handle).
       sendDraftPath = draftPath
       // caption on stdin — never in this process's argv (audit #4, war room #1/#13)
-      fileSendProc.command = ["bun", root.sendFileScript, sendChat, draftPath, "--caption-stdin"]
+      var svc2 = String(root.active.service || "iMessage")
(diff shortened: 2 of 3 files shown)


nixfred:
Thanks for the report — the double-click finding is real and worth fixing, but this can't merge as-is, and the SMS half is already shipped.

- **SMS/RCS:** 2.2.0 (3c4e65d) passes `--service` for SMS/RCS DMs on text and file sends.
- **Double-click:** `onDoublePressed` binds a signal Omarchy's `WidgetButton` does not have (checked 4.0.2 and omacom/omarchy main: only `pressed` and `wheelMoved`). On a stock shell that is a QML load error, and a load error in a bar widget silently kills BOTH surfaces (2.1.4 shipped exactly that). It only works on your patched `/usr/share/omarchy/shell`.

The mechanism you found is upstream's: `modulePointer` has no `onDoubleClicked`, so the composed double-click propagates to the center `CenterGestureArea`, which toggles transparency. The fix belongs in omacom/omarchy Bar.qml — accept `doubleClicked` on `modulePointer` when it lands on a module click target and forward it as a second press. Once that lands, Blip's existing 320 ms two-press scheme works with no change. If you open that upstream PR, link it here and I'll follow up on the Blip side if anything is still needed.

Closing this one. Please open Blip PRs from a branch rather than your fork's main so they don't drift.

PR #9, opened 2026-09-02
fix: honour EXIF orientation so iPhone photos stop rendering sideways (inline + viewer)

adamgamble:
## Problem

Photos taken on an iPhone render on their side, both inline in the conversation and in the viewer that opens on click. A camera photo stores its rotation as an EXIF orientation tag rather than in the pixels, and `sips` keeps that tag when the bridge converts HEIC to JPEG. The file that lands in `~/.cache/blip/att` is 4032×3024 landscape pixels tagged "Rotate 90 CW". Qt's `Image` ignores that tag unless asked, and imv (Omarchy's default image viewer, what `xdg-open` launches) has no EXIF support at all.

## Fix, in two commits

**1. `autoTransform: true`** on the three `Image` elements in `BlipView.qml`: the inline attachment, the link-card preview, and the avatar. Qt applies the EXIF transform at decode time, and `implicitWidth`/`implicitHeight` follow the rotated shape, so the bubble sizes correctly too.

Verified on Qt 6.11.2 (what Quickshell links against here) with a cached photo:

| | implicit size |
|---|---|
| `autoTransform: false` | 800 × 600 |
| `autoTransform: true` | 600 × 800 |

Same result on the `asynchronous: true` path, and the 800 px `sourceSize` decode bound still applies.

**2. Bake the orientation into the cached file** in `fetch.ts`, so every consumer is upright, not just Qt. A pure, bounds-checked walk of the JPEG markers reads the orientation (no pixel decode); when it isn't 1, the bytes go through `jpegtran <op> -trim -copy icc` on stdin/stdout before caching.

- Lossless, ~50 ms on a 12 MP photo.
- `-copy icc` keeps the colour profile (iPhone photos are Display P3) and drops the EXIF block, so no stale tag survives to double-rotate in Qt.
- `jpegtran` ships with libjpeg-turbo, which Qt itself links, so it is on every Quickshell box. The same library already decodes these bytes in the panel, so no new parser meets untrusted input.
- Any failure (binary missing, bad stream) keeps the original bytes; commit 1 still shows those upright inline.

Six tests: the parser in both byte orders plus truncated / out-of-range / PNG inputs, the op mapping, no spawn for upright or non-JPEG bytes, the injected-runner fetch path (args, stdin transport, cached bytes), the failure fall-back, and a real `jpegtran` run (skipped where the binary is absent) proving a tagged 16×8 becomes an upright 8×16 with no tag left.

Verified live: the same photo re-fetched as 3024×4032, no Orientation tag, Display P3 retained, 0600.

Files cached before this change keep their tag; they render upright inline (commit 1) but open sideways in imv until evicted or `~/.cache/blip/att` is cleared. Noted in the changelog entry.

Also adds an Unreleased changelog entry and an invariant note in `CLAUDE.md` so neither half gets "simplified" away later. Feel free to drop either if you'd rather own those.

https://claude.ai/code/session_01C5w1SF7gHaXPrUnoG4BCWN

### DIFF
--- a/BlipView.qml
+++ b/BlipView.qml
@@ -1263,6 +1263,7 @@ FocusScope {
                       source: root.avatarFiles[avatarCircle.avatarHandle] || ""
                       asynchronous: true
                       fillMode: Image.PreserveAspectCrop
+                      autoTransform: true
                       sourceSize.width: 96
                       sourceSize.height: 96
                       // a stale/corrupt cache file → initials, and no retry this session
@@ -1620,6 +1621,11 @@ FocusScope {
                       source: chipRow.showImage ? chipRow.fileUrl : ""
                       asynchronous: true
                       fillMode: Image.PreserveAspectFit
+                      // iPhone photos store rotation as an EXIF tag, not in
+                      // the pixels (sips keeps the tag when it converts HEIC);
+                      // Qt ignores it unless asked, so portraits came out on
+                      // their side. implicitWidth/Height follow the transform.
+                      autoTransform: true
                       // bound the DECODE in BOTH axes, not just the paint — a
                       // 12MP photo (or a 100×100000 sliver) must not cost
                       // 50 MB of texture (Codex review points 19 and #3)
@@ -1713,6 +1719,7 @@ FocusScope {
                         source: linkCard.imgUrl
                         asynchronous: true
                         fillMode: Image.PreserveAspectCrop
+                        autoTransform: true
                         sourceSize.width: 800
                         sourceSize.height: 800
                       }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,13 +2,16 @@
 
 ## Unreleased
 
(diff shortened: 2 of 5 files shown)


adamgamble:
blip-exif-repro
Here is an image to help recreate the issue. Ping me if needed.

nixfred:
Merged both halves — thank you. The second commit is the one I would have missed: the QML flag alone fixes Blip and leaves the photo sideways in imv, which is where people actually open them. I read the EXIF parser closely because it takes bytes from a stranger, and the marker walk is bounds-checked everywhere it matters. Verified on a real cached iPhone photo here: 4032x3024 tagged 'rotate 90' came back 3024x4032 with the ICC profile kept. Shipped in 2.2.3.

adamgamble:
Thank you! I had built a similar imessage clone myself, but I abandoned it
and will start sending commits to yours!

PR #10, opened 2026-09-02
fix(ui): square the bubble corner with a per-corner radius, not a stacked tail

adamgamble:
## Problem

Every received bubble that ends a run shows a lighter grey square in its bottom-left corner, where the "tail" is.

`theirsFill` is translucent (`foreground` at 14% alpha). The tail was a second `Rectangle` in `bubble.color` drawn over the corner, so the overlap composites the same translucent colour twice and reads as a lighter block. Sent bubbles never showed it because the accent fill is opaque.

## Fix

Square off the real corner with Qt 6.7+ per-corner radii on the bubble (`bottomLeftRadius` / `bottomRightRadius` to 0 on the sender's side when `groupEnd`) and drop the stacked rectangle. No overlap, so it holds for any fill alpha. One file, 6 insertions, 8 deletions.

Confirmed on screen with Quickshell 0.3.1 / Qt 6.11.2. Happy to add a changelog if needed

https://claude.ai/code/session_01C5w1SF7gHaXPrUnoG4BCWN

### DIFF
--- a/BlipView.qml
+++ b/BlipView.qml
@@ -1789,14 +1789,12 @@ FocusScope {
 
                     // iMessage squares off the corner nearest the sender on the
                     // last bubble of a run — the "tail" without drawing a tail.
-                    Rectangle {
-                      visible: modelData.groupEnd === true
-                      width: Style.space(16); height: Style.space(16)
-                      color: bubble.color
-                      anchors.bottom: parent.bottom
-                      anchors.right: bubbleRow.mine ? parent.right : undefined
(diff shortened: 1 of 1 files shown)


nixfred:
Merged — thank you. Six lines for a thing I had looked straight past: the tail was compositing the same translucent fill over itself, which is why only received bubbles showed it. Confirmed in a zoomed crop of my own panel before merging. Per-corner radius is the right primitive. Shipped in 2.2.3.

PR #11, opened 2026-09-02
Mirror pinned conversations in Blip

ezachrisen:
## Summary

- Mirror ordered pinned-conversation metadata from Messages.
- Render pinned conversations as read-only avatar/name tiles in three-column rows above ordinary conversation rows.
- Preserve existing unread behavior and activity ordering for unpinned conversations.

## Implementation

- Extend the Mac bridge chats JSON with tolerant pin-preference parsing and stable pin ordering.
- Merge pin metadata into the complete collector thread model with backward-compatible defaults and test coverage.
- Split the QML list into pinned and unpinned sections, preserving keyboard selection by chat identity.
- Document the feature and its read-only privacy behavior.

## Verification

- `bun test`: 197 passed, 0 failed.
- UI spacing and the normal Omarchy window were deployed and visually verified.

### DIFF
--- a/BarWidget.qml
+++ b/BarWidget.qml
@@ -25,7 +25,7 @@ BarWidget {
     decodeURIComponent(Qt.resolvedUrl("collector.ts").toString().replace(/^file:\/\//, ""))
 
   // ---- collector state
-  property var threads: []           // [{chat,name,handle,service,last_ts,last_text,last_from_me,count,unread}]
+  property var threads: []           // [{chat,name,handle,service,last_ts,last_text,last_from_me,count,unread,pinned,pin_order}]
   property string threadsJson: ""    // last assigned list, for no-op detection
   property int unread: 0
   property bool online: false        // the Mac is reachable
--- a/BlipView.qml
+++ b/BlipView.qml
@@ -104,6 +104,8 @@ FocusScope {
   property string pasteChat: ""
 
   readonly property var threads: hostWidget ? hostWidget.threads : []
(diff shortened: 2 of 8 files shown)


ezachrisen:
redacted-2
Screenshot of pinned conversations view

nixfred:
Merged, and thank you — this is a feature I'd wanted and hadn't got to, and the bridge/collector/QML split landed exactly where Blip keeps its logic. The screenshot helped.

I ran your bridge against my own Mac from a temp path before merging and landed three follow-ups on main in 8978a54 so the merge matches what Messages shows here:

1. **Pinned groups.** A group's pin id is its chat.db `group_id` (a dashed UUID), never the `chat_identifier` / `guid`, so both of my pinned groups were missed. `pin_order_for` now also compares `group_id` and `original_group_id`.
2. **Order.** The plist has structure: `pD.pP` is the ordered pin list and `pD.pZ` maps a group's pin id to `{ "o": <original group id> }`. Flattening every string ranked one group at position 2 (its `pZ` alias) instead of 8. The reader now walks `pD.pP` with the `pZ` aliases and keeps your flatten as the fallback for an unknown layout.
3. **Keyboard.** `moveCursor` walks `root.threads`, and pinned threads sort first in it, so j/k had no visible position on the tiles for those presses. The tile now highlights when the cursor is on it.

One thing I noticed that is *not* yours: a group that migrated between services has two `chat` rows and shows twice, now as two tiles. `chats` needs the same one-row-per-group dedupe #8 gave `groups`; it's on the ROADMAP.

Ships in the next release with the SMS fix.

PR #12, opened 2026-09-03
feat(app): live search and Super+M keys

johanthoren:
## Why

The Super+M window ignored `n`, `/`, and Esc that the bar popout already handles. New-message and message search waited for Enter. A lowercase query missed newer iMessage bodies that live only in `attributedBody`, so May hits could outrank August ones.

This change makes the app window match the popout keys, searches as you type, puts conversations above messages, and ranks whole-word hits by the message timestamp.

## Scope

Ten commits on `fix/app-window-n-slash`, rebased onto current `upstream/main` (`81fbead`).

- `BlipWindow.qml` and `BlipView.qml` catch `n`, `/`, and Esc. `unwind()` still closes the share sheet first.
- The new-message field. Arrow keys and Enter pick a handle. `contact-search.ts` ranks by thread `last_ts`.
- Contact dump is cached for 60s in `$XDG_RUNTIME_DIR` and filtered with `fuzzyScore`. The field is polled while `newMode` is true because `TextField` `textChanged` did not fire in the window.
- `/` search matches sidebar conversations first (`matchConversations` in `search.ts`), then `imsg search` hits.
- `messageMatchScore` ranks case-insensitive whole words above substrings, then message `ts`.
- `bridge/mac/imsg` `cmd_search` scans lower, upper, and title-case needles in `attributedBody`.
- `CHANGELOG.md` Unreleased notes the work.

Out of scope. Copy the updated `imsg` to `~/.blip/bin` on the Mac gateway or lowercase search still misses capitalized attributedBody. Number-row jumps are a follow-up.

## Tradeoffs

QML duplicates `fuzzyScore` for the instant conversation preview. The thread list is already in memory. Sending `last_text` on argv would put message bodies on the command line. After bun returns, `runSearch` is the ranking source of truth.

The new-message field is polled every 50ms instead of a `textChanged` handler. That is the path that ran in Super+M.

One PR with ordered commits instead of a fork stack. Each commit is reviewable on its own.

## Blast Radius

Super+M and the bar popout share `BlipView.qml`. Search, new-message, and Esc paths change. Sends, reads, and the collector do not.

The Mac `imsg` search SQL ORs extra `instr` needles. Install the updated binary on the gateway for the case-fold to apply.

## Verification

- `bun test` at `59dd12d`. 249 pass, 0 fail.
- Live on the Super+M window. `n` then type finds contacts without Enter. `/` then type lists conversations then messages. After the Mac `imsg` deploy, a lowercase whole-word query matches capitalized bodies and ranks by message time.
- QML contracts in `ui.test.ts` for catch helpers, live contact search, and people-first search.
- Ranking tests in `tier2.test.ts` use Alice, Bob, and English phrases only.


nixfred:
Merged — thank you. Ten ordered commits with a blast-radius section made this genuinely reviewable, which is not the norm for a 600-line change.

Two notes for the record.

One thing you fixed without knowing it: removing that `content.forceLayout()` call is correct beyond the reason given. That call has been **throwing on every invocation** — `TypeError: Property 'forceLayout' of object QQuickColumnLayout is not a function` — which I saw in my own logs before this PR existed. The comment above it was describing a fix that had stopped running.

One change I made on merge: the thread-recency map now travels on stdin rather than argv (`--recency-stdin`, follow-up commit 8f0c7a1). It is every handle you have a conversation with, and argv is readable by any process on the machine through `ps` — the same reason message text has never gone that way. You had already kept `last_text` off the command line for exactly that reason, so this is that instinct applied one field further. Ranking behaviour is unchanged.

PR #13, opened 2026-09-03
feat(app): jump to a thread with 1-9

johanthoren:
## Why

After Super+M, the sidebar is already numbered in your head. macOS Messages uses Super+number. Omarchy already owns Super+1-9 for workspaces, so this is a bare digit, the same as `n` and `/`.

## Scope

Depends on #12 (`fix/app-window-n-slash`). Unique commit: `87a1357`. Unique diff: https://github.com/johanthoren/blip/compare/fix/app-window-n-slash...feat/thread-number-hints

Until #12 merges, GitHub will also list that PR's commits against `main`. Review `87a1357` for this change.

- Super+M then `1`-`9` opens that sidebar thread without Enter. Pins take those slots first.
- The first nine rows show the digit.
- Empty compose jumps. Non-empty compose, search, new-message, a queued file, leftover search, and Shift+digit stay in the field.
- A digit with no matching row is typed, not eaten.

## Tradeoffs

The window uses `catchNavText` (no `PanelKeyCatcher`). The popout still jumps `1`-`9` only from the list, because compose focus blocks the catcher.

## Blast Radius

`BlipView.qml` key catch and sidebar rows. Sends, search ranking, and the collector do not change.

## Verification

- `bun test` at `87a1357`. 250 pass, 0 fail.
- Live Super+M window. Digit hints on the first nine rows. `3` opened the third row. `1` jumped from an already-open empty compose.
- `ui.test.ts` locks `event.text` before `Key_1`, `draftPath` stand-down, and `catchNavText` returning only when `handleTextKey` actually opens a thread.


nixfred:
Merged — thank you. A bare digit rather than SUPER+digit because Omarchy owns SUPER+1-9 is the right read of the platform, and the stand-down list (non-empty compose, search, new-message, a queued file, Shift+digit, and a digit with no matching row) is the part that makes it safe to live with. Digit hints are on the pinned tiles on my screen now. Merged by hand for the same one-hunk CHANGELOG conflict as #12.

PR #14, opened 2026-09-03
imsg: hide announcement rows, they are not messages

Fileri:
## What

`imsg` no longer returns Messages.app's own announcement rows — someone joined or left a group, a rename, location sharing started or stopped (`item_type != 0`). They are hidden at the source, next to Recently Deleted, so every query agrees: `recent`, `thread`, `chats`, `groups`, and the unread ledger the collector builds from them.

User-visible effect: no more empty bubbles that count as unread forever and pull a conversation to the top. Mac side needs the install one-liner re-run so `~/.blip/bin/imsg` picks it up (noted in CHANGELOG).

## Why

Apple writes these notices into the message table with an empty body and never marks them read. The bridge passed them through, so Blip treated each one as an unread inbound message. An iCloud re-sync after a restart lands several at once: on my Mac, five location-sharing notices arrived in the same second and lit two family threads with "two new messages" that did not exist. The database had 42 such rows sitting at `is_read = 0`, some from 2025.

Rendering notices the way Messages does, as grey centered text, would be a follow-up: the bridge would emit them tagged, and the collector would keep them out of the unread ledger and toasts. Out of scope here.

## How it was verified

- [x] `bun test` is green (CI will check) — 234 pass, plus `py_compile` on every Mac tool
  - this change does not send; it only filters reads
  - no UI change; the bubbles simply stop appearing

Against a live `chat.db`, old tool vs new, compared by row id and counts only:

- `recent 60`: exactly the five `item_type = 4` rows left the window; five older real messages took their place
- no `item_type != 0` row remains in the newest 500
- `chats`: 30 before, 30 after · `groups`: 149 before and after, 4 named before and after (rename-row name recovery reads the raw table and is unaffected)
- unread inbound rows in the newest 500: 12 → 2
- `IMSG_INCLUDE_DELETED=1` still works; the macOS 12 fallback (no Recently Deleted table) now filters announcements too instead of reading the raw table

Invariants touched:
- **Deleted messages stay in chat.db for 30 days** — same mechanism, unchanged behaviour; the filter moved into `message_src()`
- **New: Announcements are not messages** — added to CLAUDE.md, with a matching design note in README next to "Deleted means deleted"

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 ## Unreleased
 
+- **Location-sharing notices no longer show up as empty unread messages.**
+  Messages.app writes its own announcements — someone joined or left, a
+  rename, location sharing started or stopped — into the message table with
+  `item_type != 0`, and never marks them read. The bridge passed them through,
+  so each one became an empty bubble that counted as unread forever and pulled
+  its conversation to the top; an iCloud re-sync after a restart lands several
+  at once. `imsg` now hides them at the source, next to Recently Deleted, so
+  every query agrees. Group names are unaffected: rename rows are still read
+  for that. Mac side: re-run the install one-liner so `~/.blip/bin/imsg` picks
+  it up.
 - **Reads now reach the Mac, and your phone.** Blip's "mark all read" cleared
   the badge on Linux and nothing else; the iPhone kept its red dots. The old
   note called this impossible because `open imessage://<handle>` does not flip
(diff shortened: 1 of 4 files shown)


nixfred:
Merged — thank you, and this one is bigger than your own repro suggested. You found 42; my Mac has **811** of these sitting at `is_read = 0`, the oldest from 2021-09-14, and not one of them carries any text. Every one was a phantom unread that could pull a conversation to the top. Filtering at the source so `recent`, `thread`, `chats`, `groups` and the unread ledger all agree is exactly right. Shipped and deployed to my gateway.

PR #15, opened 2026-09-03
Clock and dates follow Qt format strings on the widget entry

Fileri:
## What

Bubble timestamps, read receipts, day dividers and the conversation list now follow three Qt format strings on the widget's shell.json entry — the same inline-settings shape Omarchy's clock uses for its `format`:

```jsonc
{ "id": "nixfred.blip", "timeFormat": "HH:mm", "dateFormat": "dd.MM", "dateFormatWithYear": "dd.MM.yyyy" }
```

Unset, `timeFormat` follows the locale (`h:mm AP` when the locale's short time has an AM/PM marker, else `HH:mm`) and the date patterns stay what Blip shows today (`MMM d`, `MMM d, yyyy`). A US install keeps its bubbles as they were; the one visible change there is the list, where an older row now reads `Sep 2 11:59 PM` instead of the raw `09-02 23:59` slice.

| defaults (US locale) | `HH:mm` · `dd.MM` · `dd.MM.yyyy` |
|---|---|
| defaults | configured |

Both rendered by `scripts/demo/blip-shots` against the fake bridge — invented people, real code paths.

## Why

Two clocks in one window, and no setting for either. `thread.ts` hardcoded 12-hour for bubbles and receipts; `BlipView.fmtTime` showed the list in 24-hour by slicing the raw stamp. Next to an Omarchy bar clock set to `dddd HH:mm`, a European install read 21:08 in the bar and the list, and 9:08 PM in the conversation.

Omarchy already settled how a widget takes this kind of preference: a Qt format string inline on the entry, read with `setting()`. Copying that is smaller and more predictable than a Blip-only 12h/24h switch, and it covers dates in the same stroke.

## How

- **`thread.ts`** — `Formats { time, date, dateWithYear }` with the Messages-style defaults. `formatsFromArgv()` reads `--time-format`, `--date-format`, `--date-format-with-year`; an empty value means the default. `formatStamp(ts, pattern)` implements the Qt tokens Blip needs (`d dd ddd dddd · M MM MMM MMMM · yy yyyy · h hh H HH · m mm · s ss · AP ap A a`, quoted literals, `''` as an apostrophe); `h` is 12-hour only when an AM/PM token is present, as in Qt. It works on the string, never a `Date`, so a Mac-local stamp still never meets the Linux time zone. The hand-rolled 12-hour conversion and the month table are gone. `clockLabel`, `dayLabel`, `receiptLabel`, `decorate` and `loadThread` take the formats, defaulting to today's behaviour.
- **`BarWidget.qml`** — three `readonly` properties via `setting()`; `localeTimeFormat` is the locale-derived default.
- **`BlipView.qml`** — takes the strings from the host, passes the three flags to `thread.ts`, and formats the list with `Qt.formatTime` / `Qt.formatDate`. No arithmetic in QML.
- **`scripts/demo/DemoShell.qml`** — the screenshot harness's stand-in host supplies the three too (env-overridable), which is how the shots above were made.
- **README** (install step 4) and **CHANGELOG** (Unreleased).

`Today` and `Yesterday` stay words, as in Messages. Names and markers are English in both formatters — the same choice Omarchy's calendar makes by pinning its labels to `en_US`.

## How it was verified

- [x] `bun test` is green (CI will check) — 259 pass; 21 new tests, including a 17-row pattern table for `formatStamp`
- [x] Any send/receive behavior was exercised against **my own number only** — this change does not send; `thread.ts` was run against my own thread through the live bridge, with the flags, with empty flags, and without
- [x] UI change → screenshots above (harness, invented people). On a live Omarchy shell: restart clean, no QML warnings, list and thread agree
- [ ] Touches an invariant in `CLAUDE.md` → none touched

Cross-checks on a live Omarchy box:
- `formatStamp` vs `Qt.formatDateTime` on 23 patterns (including `h 'o''clock'`, `'at' h:mm ap`, `ddd d MMM`): identical
- Locale default under 12 locales (`en_US en_GB en_AU nb_NO de_DE fr_FR es_MX hi_IN ko_KR ja_JP ar_EG C`), each in its own Qt process: `h:mm AP` for the AM/PM locales, `HH:mm` for the rest — including `ko_KR`'s `Ap h:mm` and `ja_JP`'s `H:mm`. With an explicit pattern Qt keeps English names and markers in every one of them, so the list and the bubbles never diverge.

### DIFF
--- a/BarWidget.qml
+++ b/BarWidget.qml
@@ -24,6 +24,18 @@ BarWidget {
   readonly property string collectorPath:
     decodeURIComponent(Qt.resolvedUrl("collector.ts").toString().replace(/^file:\/\//, ""))
 
+  // Clock and date patterns from this widget's shell.json entry — Qt format
+  // strings, exactly as Omarchy's clock takes `format`. Qt formats the list
+  // with them and thread.ts formats the bubbles with the same strings, so the
+  // two never disagree. Unset, the time follows the locale (an AM/PM marker
+  // means 12-hour) and dates read the way Messages writes them.
+  readonly property string localeTimeFormat:
+    /ap/i.test(Qt.locale().timeFormat(Locale.ShortFormat)) ? "h:mm AP" : "HH:mm"
+  readonly property string timeFormat: formatSetting("timeFormat", localeTimeFormat)
+  readonly property string dateFormat: formatSetting("dateFormat", "MMM d")
+  readonly property string dateFormatWithYear: formatSetting("dateFormatWithYear", "MMM d, yyyy")
+  function formatSetting(name, fallback) { var v = String(setting(name, "")); return v === "" ? fallback : v }
+
   // ---- collector state
   property var threads: []           // [{chat,name,handle,service,last_ts,last_text,last_from_me,count,unread,pinned,pin_order}]
   property string threadsJson: ""    // last assigned list, for no-op detection
--- a/BlipView.qml
+++ b/BlipView.qml
@@ -22,6 +22,11 @@ FocusScope {
 
   // ---- host contract (docs/app-design-review.md) ----------------------
   property var hostWidget: null
+  /** Qt format strings, owned by the host widget (see BarWidget). Empty when no host is
+   *  attached or the host has none, which thread.ts and Qt both read as "use the defaults". */
+  readonly property string timeFormat: (hostWidget && hostWidget.timeFormat) || ""
+  readonly property string dateFormat: (hostWidget && hostWidget.dateFormat) || ""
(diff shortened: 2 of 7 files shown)


nixfred:
Merged — thank you. Two clocks in one window was a fair description and I had stopped seeing it. Following Omarchy's own inline-settings shape instead of inventing a Blip-only toggle is the part I appreciate most; it means there is nothing new for anyone to learn. Merged by hand for a one-hunk CHANGELOG conflict after #14 landed, nothing else. And thank you for rendering the before/after with `scripts/demo/blip-shots` — that harness went in a day ago and you are the first person to use it, which is exactly what it was for.

PR #19, opened 2026-09-03
bridge: ignore Family Sharing child-delegate Contacts stores when resolving names

adamgamble:
## Problem

With Family Sharing's **Manage Contacts** (Screen Time), each child's address book is mirrored onto the parent's Mac as its own CardDAV store under `~/Library/Application Support/AddressBook/Sources/<id>/`. Contacts.app hides those stores from *All Contacts*, but the bridge globbed every `.abcddb` on disk and let each one vote on a name. With two sons, the number saved as **Monica Gamble** in my own contacts rendered as **Mom** in every Blip thread, pinned tile, and group name — two votes to one.

## Fix

The Accounts database (`~/Library/Accounts/Accounts4.sqlite`, readable under the same Full Disk Access grant as `chat.db`) flags those accounts with the property `isChildDelegate = YES`, and the `Sources/<id>/` folder name is the account identifier. `imsg`, `contacts`, and `blip-check` now read that flag and drop those folders before building the name index, looking up avatars, or merging contacts.

- Unreadable Accounts database → empty skip set → exactly the previous behavior.
- `blip-check` reports `N Family Sharing child list(s) ignored`.
- CLAUDE.md gains the exception under the cross-source invariant; CHANGELOG has the user-facing note (Mac side needs the install one-liner re-run).

## Verified

macOS 26.6, live bridge:

```
skip: ['84009003', 'B8EBFFFC']
sources used: ['50B7BBAE']
+12707845266 -> Monica Gamble
  ✅  contacts   326 contacts readable (2/2 sources, 2 Family Sharing child list(s) ignored)
```

`bun collector.ts --deep` now names the DM and both groups "Monica Gamble, …". `bun test`: 279 pass. CI's `py_compile` covers the three edited tools.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 ## Unreleased
 
+- **Your kids' contact lists no longer rename your contacts.** Family
+  Sharing's "Manage Contacts" (Screen Time) mirrors each child's address book
+  onto the parent's Mac as a separate Contacts store. Contacts.app keeps those
+  out of All Contacts, but the bridge read every store on disk and let them
+  vote on names — so with two sons, the number saved as "Monica Gamble" in
+  your own contacts showed as "Mom" in every Blip thread, tile and group
+  name, two votes to one. `imsg`, `contacts` and `blip-check` now skip stores
+  whose account is flagged `isChildDelegate` in the Accounts database; a Mac
+  where that database cannot be read behaves exactly as before. `blip-check`
+  says how many child lists it ignored. Mac side: re-run the install
+  one-liner. Names and photos that exist ONLY in a child's list fall back to
+  the bare number, which is what Contacts.app shows you too.
 - **Location-sharing notices no longer show up as empty unread messages.**
   Messages.app writes its own announcements — someone joined or left, a
   rename, location sharing started or stopped — into the message table with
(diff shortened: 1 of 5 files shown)


nixfred:
Merged in 1903e24 — thank you. This is the right shape of fix: you did not touch the majority-wins rule, you carved out the stores Contacts.app itself does not count. That distinction is what made it easy to accept.

I verified it on my own Mac, and the result is worth writing down for anyone who finds this later.

I have three AddressBook sources — two of them extra, holding 2217 and 2485 contacts against my own 2217. If this ever misfired here it would mis-resolve essentially every name I have. It does not, and the reason is the part of your patch that looks least important:

```
skip set: set()
sources used: ['8F8569EB', 'DBC79919', 'E7090AC8']
contacts   13511 readable (4/4 sources)   ← unchanged
```

Both of my accounts **carry an `isChildDelegate` property and both archive `False`.** So the property exists on Macs with no Family Sharing child contacts at all. A presence check — `WHERE ZKEY = 'isChildDelegate'` and skip — would have silently dropped two thirds of my address book and looked like a Contacts sync bug. `_archived_bool()` decoding the NSKeyedArchiver value is what makes this safe, and I would not have guessed that mattered until I ran it.

Also appreciated: the empty-set-on-failure path, so an unreadable Accounts database means the old behaviour rather than no contacts, and `blip-check` staying quiet when the count is zero instead of reporting a confusing "0 ignored".

Merged by hand for a one-hunk CHANGELOG conflict, nothing else. Deployed to my gateway; `blip-check` reads 13511 contacts across 4/4 sources, unchanged, which is exactly what a correct no-op looks like here.

That is three from you now — the EXIF orientation, the bubble corner, and this. Thank you.

PR #20, opened 2026-09-03
Contact management: review, compare, edit, merge, and clean up Mac contacts from Blip

jondkinney:
This lets Blip manage Mac contacts from Linux. My address book had accumulated years of duplicates across iCloud, Gmail, Exchange, and On My Mac, and fixing them meant walking over to the Mac — so now Blip can compare cards side by side across accounts, edit them, consolidate duplicates into a chosen surviving card, merge two differently-named cards when someone changes their name, and run a read-only scan that finds the cleanup candidates for you.

Getting writes to actually work was most of the effort. macOS 15.5 has three broken Contacts scripting verbs (add-to-person fails with "No error. (0)", removals with "Message not understood", and person deletion just doesn't), and one damaged stored row makes CNContactStore fail whole saves with a persistent Cocoa error 134092. The bridge works around all of it: unchanged labeled values are reused instead of rebuilt so contactsd never faults the bad rows, every mutating stage retries on a fresh store session, and when the native path still can't save, the same reviewed plan goes through Contacts.app's own write path. Cleanup scans are cached against a per-account store fingerprint, so re-opening the page answers in ~0.1s when nothing changed on the Mac instead of re-scanning 19k cards.

The safety rails follow the rules already in CLAUDE.md: handles travel over stdin and never argv, everything crossing ssh is bounded JSON, every mutation shows a hashed preview that's revalidated at apply time, destructive saves write an undo receipt first, session-only candidate selection stays separate from saved display rules in `identities.json`, and Blip never touches the AddressBook SQLite database. Cross-name merges require the user to explicitly present both cards — no guessing that two names are the same person.

Tests: 350 in `bun test`, 41 in `bridge/mac/contacts_test.py`, plus `imsg_test.py`. I've been running this against my real contacts for a while now — the workarounds above all came from cards in my own address book that refused to save.

Independent of #21 — either can land alone. They share the preferences groundwork and both touch BlipView/BlipSettings, so whichever lands second gets an immediate rebase from me.

### DIFF
--- a/BarWidget.qml
+++ b/BarWidget.qml
@@ -24,14 +24,22 @@ BarWidget {
   readonly property string collectorPath:
     decodeURIComponent(Qt.resolvedUrl("collector.ts").toString().replace(/^file:\/\//, ""))
 
+  // One validated preference source for both render surfaces. The popout and
+  // app update together, and external dotfile restores are picked up live.
+  BlipPreferences { id: preferences }
+
   // Clock and date patterns from this widget's shell.json entry — Qt format
   // strings, exactly as Omarchy's clock takes `format`. Qt formats the list
   // with them and thread.ts formats the bubbles with the same strings, so the
-  // two never disagree. Unset, the time follows the locale (an AM/PM marker
-  // means 12-hour) and dates read the way Messages writes them.
+  // two never disagree. An explicit shell.json format wins; otherwise the
+  // portable Blip preference decides 12- vs 24-hour, and an unloaded
+  // preference file falls back to the locale (an AM/PM marker = 12-hour).
   readonly property string localeTimeFormat:
     /ap/i.test(Qt.locale().timeFormat(Locale.ShortFormat)) ? "h:mm AP" : "HH:mm"
-  readonly property string timeFormat: formatSetting("timeFormat", localeTimeFormat)
+  readonly property string preferredTimeFormat: preferences.loaded
+    ? (preferences.use12HourConversationTimes ? "h:mm AP" : "HH:mm")
+    : localeTimeFormat
+  readonly property string timeFormat: formatSetting("timeFormat", preferredTimeFormat)
   readonly property string dateFormat: formatSetting("dateFormat", "MMM d")
   readonly property string dateFormatWithYear: formatSetting("dateFormatWithYear", "MMM d, yyyy")
   function formatSetting(name, fallback) { var v = String(setting(name, "")); return v === "" ? fallback : v }
@@ -112,6 +120,7 @@ BarWidget {
     target.bar = root.bar
     target.anchorItem = button
     target.hostWidget = root
+    target.preferences = preferences
   }
   onBarChanged: injectPanel()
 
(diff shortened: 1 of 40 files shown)


nixfred (changes requested):
This is a serious piece of engineering and I want to say that before I say anything else, because what follows is a request to split it rather than a criticism of it.

**What I checked, and what I got wrong first.** My initial worry was that a Linux box holding the confined ssh key could now delete contacts on the Mac. I went looking for a client-side-only toggle and expected to find one. Instead: `require_write_gate()` on the Mac, and two independent opt-ins — `contact_writes=on` in the owner's Linux config **and** a 0600 `~/.blip/contact-writes-enabled` file on the Mac **that the forced-command key cannot create**. That is the right posture and better than I assumed. Routing every mutation through the existing `contacts` verb so `blip-dispatch` stays at five tools was also the correct instinct.

The macOS 15.5 workarounds are the part I would never have found on my own: three broken Contacts scripting verbs, and one damaged stored row making `CNContactStore` fail whole saves with 134092. Reusing unchanged labeled values so `contactsd` never faults the bad rows is a real piece of archaeology.

**Why I still want it split.**

1. **Scope.** Blip is an iMessage client. This makes it a contacts manager as well, and that is a direction decision for the project rather than a code review. It deserves its own conversation, not a merge button at the end of a 14,600-line diff.

2. **The install cost falls on everyone.** `install.sh` now hard-fails without `swiftc`:

   ```
   install.sh: swiftc not found — install Xcode Command Line Tools
   exit 1
   ```

   That makes a multi-gigabyte Xcode dependency mandatory for someone who only wants to read iMessage on Linux. Even if the contacts feature lands, that check should skip the Swift helper and continue, leaving contact writes unavailable, rather than stopping the install.

3. **The read-only half stands on its own and I want it.** The duplicate scan across iCloud/Gmail/Exchange/On My Mac, the side-by-side compare, and the per-account fingerprint cache that answers in ~0.1s instead of re-scanning 19k cards — that is useful, carries no write risk, needs no Swift, and I would review it quickly. **Send that as its own PR and it gets a fast yes.** Then the write path as a third PR, judged on its own merits, with the install made conditional.

**Ordering:** this and #21 touch 19 of the same files. I have asked there for the rendering fixes to come out first as a small PR, then settings, then this. `main` also moved today — 2.3.0 plus PRs from five other people — so a rebase is coming regardless, and I would rather you spend it once, on a smaller diff.

Genuinely: thank you. Nothing here reads as careless, which is why I would rather land it in pieces I can defend than in one I cannot.

jondkinney:
Agreed on the split. Coming next as its own PR: the read-only half — the cross-source duplicate scan, side-by-side compare, and the per-account fingerprint cache — with no write path and no Swift dependency. The write path will follow separately, with install.sh degrading gracefully when swiftc is absent (skip the helper, leave contact writes unavailable) rather than hard-failing. Ordering per your note in #21: after the rendering fixes and the settings conversation.

jondkinney:
Closing in favor of the split you asked for: the read-only half (cross-source duplicate scan, fingerprint cache, card review — no Swift, no writes) is #24, and the write path (both gates, previews, undo receipts, compare, vCard export) is #25, sitting in draft behind it with install.sh made conditional on swiftc as requested — it skips the compiled helper and continues, leaving writes unavailable, instead of hard-failing. Nothing from this PR is lost between the two; the branch stays on my fork if anything needs cross-referencing.

PR #21, opened 2026-09-03
Appearance settings, correct-scale rendering, and popout polish

jondkinney:
This adds a settings surface to Blip and then uses it: a portable preferences file with a live appearance editor, plus a pile of rendering fixes I've been accumulating while using the app daily.

The preferences live in a plain JSON file at `~/.config/blip/preferences.json` so they can ride along in dotfiles. `preferences.ts` owns the schema and all the I/O — it refuses symlinks, wrong owners, and malformed or oversized files, and writes atomically at 0600 — so QML never touches the writable file directly. The editor covers bubble colors, window opacity, font scale, density, sidebar width, avatar size, corner roundness, and a 12/24-hour time preference that feeds the timeFormat plumbing you added in 2.3.0. The fun part is the preview: it's a true miniature of the app laid out at real dimensions and shrunk with one scale transform, and the page chrome freezes while you drag a slider so only the preview moves. The popout widens while settings are showing so the preview can sit beside the controls.

The rendering side fixes things you mostly notice on a HiDPI screen: Retina screenshots were drawing at 2x their intended size (fetch.ts now reads the image header and honors pixel density), inline media caps in logical units, message links take the bubble's text color instead of theme blue, and grouped bubble corners draw as one shape instead of overlapping darker. Conversations also pick up some Messages details — tall link artwork keeps its aspect ratio, tapbacks land on unfurled cards too, one-to-three-emoji messages render big and bubble-free, incoming group runs show the sender's avatar, and a message that's just the URL of its own link card gets suppressed. Sidebar: hover now fills the full row, separators hide under the hovered row, pinned tiles center properly, and the 1-9 jump digits from 2.3.0 moved into the chronological rows so there's a single pinned grid. Renamed/migrated group chats coalesce into one conversation when the title and participant set agree.

326 tests pass, and `python3 bridge/mac/imsg_test.py` covers the chat projection against a fixture chat.db.

This is independent of #20 — either can land alone. They both touch BlipView and BlipSettings, so whichever goes in second will need a rebase; happy to do that immediately once one lands.

### DIFF
--- a/BarWidget.qml
+++ b/BarWidget.qml
@@ -24,14 +24,22 @@ BarWidget {
   readonly property string collectorPath:
     decodeURIComponent(Qt.resolvedUrl("collector.ts").toString().replace(/^file:\/\//, ""))
 
+  // One validated preference source for both render surfaces. The popout and
+  // app update together, and external dotfile restores are picked up live.
+  BlipPreferences { id: preferences }
+
   // Clock and date patterns from this widget's shell.json entry — Qt format
   // strings, exactly as Omarchy's clock takes `format`. Qt formats the list
   // with them and thread.ts formats the bubbles with the same strings, so the
-  // two never disagree. Unset, the time follows the locale (an AM/PM marker
-  // means 12-hour) and dates read the way Messages writes them.
+  // two never disagree. An explicit shell.json format wins; otherwise the
+  // portable Blip preference decides 12- vs 24-hour, and an unloaded
+  // preference file falls back to the locale (an AM/PM marker = 12-hour).
   readonly property string localeTimeFormat:
     /ap/i.test(Qt.locale().timeFormat(Locale.ShortFormat)) ? "h:mm AP" : "HH:mm"
-  readonly property string timeFormat: formatSetting("timeFormat", localeTimeFormat)
+  readonly property string preferredTimeFormat: preferences.loaded
+    ? (preferences.use12HourConversationTimes ? "h:mm AP" : "HH:mm")
+    : localeTimeFormat
+  readonly property string timeFormat: formatSetting("timeFormat", preferredTimeFormat)
   readonly property string dateFormat: formatSetting("dateFormat", "MMM d")
   readonly property string dateFormatWithYear: formatSetting("dateFormatWithYear", "MMM d, yyyy")
   function formatSetting(name, fallback) { var v = String(setting(name, "")); return v === "" ? fallback : v }
@@ -112,6 +120,7 @@ BarWidget {
     target.bar = root.bar
     target.anchorItem = button
     target.hostWidget = root
+    target.preferences = preferences
   }
   onBarChanged: injectPanel()
 
@@ -137,7 +146,10 @@ BarWidget {
     id: windowLoader
     active: root.leader                // created at start so it can self-restore
     source: Qt.resolvedUrl("BlipWindow.qml")
-    onLoaded: item.hostWidget = root
+    onLoaded: {
+      item.hostWidget = root
+      item.preferences = preferences
+    }
   }
   readonly property bool windowVisible: windowLoader.item ? windowLoader.item.visible === true : false
   function hideWindow() {
@@ -174,6 +186,14 @@ BarWidget {
     Quickshell.execDetached(["sh", "-c",
       'for i in 1 2 3 4 5 6 7 8; do a=$(hyprctl clients -j | jq -r \'.[] | select(.title | startswith("Blip")) | .address\' | head -1); [ -n "$a" ] && break; sleep 0.15; done; ' +
       '[ -n "$a" ] && hyprctl dispatch "hl.dsp.focus({ window = \\"address:$a\\" })" >/dev/null'])
+  }
+  function showSettings(page) {
+    showApp()
+    Qt.callLater(function() {
+      if (!windowLoader.item) return
+      windowLoader.item.visible = true
+      windowLoader.item.openSettings(page)
+    })
   }
   /** Either surface open → keep the deep (complete) thread list. */
   function anySurfaceOpen() {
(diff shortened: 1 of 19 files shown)


nixfred (changes requested):
Thank you for this — and for #20 alongside it. Two PRs of this size in one evening, both with a real writeup, is not a normal drive-by and I read them properly.

I want to take this in two pieces rather than one, and I want to be clear that is about sequencing, not quality.

**The rendering fixes I would take today.** The Retina 2x bug in particular is a genuine find — `fetch.ts` reading the image header and honouring pixel density is exactly right, and I would not have gone looking for it. Same for links taking the bubble's text colour instead of theme blue, grouped corners drawing as one shape instead of overlapping darker, and tall link artwork keeping its aspect ratio. Those are unambiguous, they are small, and they need no product decision from me. **Please send them as their own PR against `main` and I will merge it.**

**The settings surface is a separate call.** `preferences.ts` is careful work — refusing symlinks and wrong owners, atomic 0600 writes, QML never touching the file — and the miniature live preview is a genuinely nice idea. But it is about a thousand lines of new product surface, and Blip has so far deliberately taken its look from Omarchy's theme rather than owning a settings page. That is a direction question, and it deserves to be judged on its own rather than riding in with a set of bug fixes I obviously want.

**Ordering, since this and #20 touch 19 of the same files** including `BlipView.qml`, `BlipSettings.qml` and `preferences.ts`: whichever lands second needs a rebase, and I would rather tell you the order than resolve a nineteen-file overlap inside your code on your behalf. My order is: rendering fixes first (small, no decision), then we talk about settings, then the contacts work in #20 last.

One practical note that applies to both: `main` moved today — 2.3.0 shipped, plus PRs from five other people — so a rebase is coming either way. Sorry about the timing.

jondkinney:
Thanks for the thoughtful review — agreed on all counts. The three rendering fixes (plus the ESM test-loader fix) are extracted and rebased onto current main as #23; the grouped-corner fix turned out to already be solved on main by #10's per-corner radii, so #23 notes that and omits it. I'll rebase this PR down to the settings surface only once #23 lands, and happy to have the direction conversation on the settings question whenever suits — no urgency from my side.

jondkinney:
On the settings question:

The page doesn't change how Omarchy themes Blip. It changes how Blip renders inside the theme. The theme answers palette and type — accent, foreground, font — and the page keeps that as the source of truth. Bubble colors default to "theme" and follow the accent, and every control defaults to a visual no-op, so anyone who never opens the page gets exactly today's Blip.

What the theme can't answer is density, avatar size, corner rounding, background opacity, font scale, sidebar proportions. There's nothing in an Omarchy theme to derive those from. They depend on display scale, eyesight, and taste — I move between a 1x and a 2x monitor daily and want different density on each. 2.3.1 adjusted the same class of values by hand: sidebar width 440 → 352, avatar 30 → 34, one-line previews to two. Those are good defaults and I kept them, but they're judgment calls, and this is the set of values where people will reasonably want different answers.

On config keys vs a GUI: ui_font=theme and the Qt time-format strings are the right pattern for discrete choices and I'd keep using them for those. These are continuous values. Nobody finds density 0.9 vs 1.05 by editing JSON and reloading the shell each round trip. The live preview is what makes tuning them practical, and the result still lands in a portable owner-only preferences.json (atomic, 0600, QML never parses the writable file), so dotfiles users end up with the same artifact a config key would produce.

Scope: the page stays appearance-only. Discrete and behavioral options keep going through bridge.conf and shell.json as they do now.

If any individual control feels like too much, I'll drop it.

nixfred:
Thank you for this, and for the patience across three rounds — extracting #23 the way you did was exactly right, and it shipped: the Retina-scale media, the readable link colors and the natural link artwork are all live in 2.3.x and they made a visible difference on my own machine.

I have to close this one, and I owe you the real reasoning rather than a shrug.

The implementation is not the problem. Atomic writes, 0600, ownership and size validation on the preferences file, the number rules with explicit min/max/digits — that is more careful than most of what I write, and your argument in the thread is a good one: the theme answers palette and type, and your page only changes how Blip renders inside it. I read it properly and it is not wrong.

The decision went the other way on shape, for three reasons:

1. **Blip is aiming at one opinionated look.** The whole recent direction has been to make it read as iMessage, and the UI has no unit tests — every layout bug on this project so far was found by taking a screenshot and looking at it. Eleven tunable knobs multiply the states a bug can hide in, on the surface where we can least afford that.

2. **It duplicates a mechanism that already exists.** `fontScale` is `ui_font_size`, which landed the same day as a `bridge.conf` key (#30). Blip already has exactly one config surface and it is that file.

3. **`hideShortCodeConversations` defaults to `true`.** This is the one that decided it. Short-code SMS is how my power company reaches me, and getting a message from 99123 to arrive at all took a full day of work — short-code threads were loading empty because the chat id needed a `+` prefix. A default that hides messages is not a preference, and I would not have caught this by reading the diff; I caught it by recognising the number.

So the standing rule is now written into CLAUDE.md: configuration is `bridge.conf` keys, and anything worth configuring becomes another key. If there is a specific setting you actually want — background opacity is the one I would argue for myself — open a small PR that adds it as a key and I will take it seriously and quickly.

None of your work here is wasted from my side: the rendering half is already in, and #24 is the piece I still want. Details there in a moment.

jondkinney:
Sounds good @nixfred, I can work on some tweaks to bring over the transparency one. Without that it makes reading the text a bit difficult. Especially because the existing default transparency doesn't seem like the correct default amount of opaqueness. I think it's more translucent than other apps are by default. Though I could be wrong!

In terms of the short code hiding - that was only for the contact review screen (since you probably don't store your short codes as contacts? Maybe you do...I don't), But it didn't hide the short code messages in the main UI. At least, that wasn't the intention. If it actually did that, then that was a bug/oversight on my part. So you may want to still consider hiding short codes for the contact review half of the work that's landing? LMK. 

Lastly, the ability to resize the message list bar width I think it useful because one of my pinned groups is my two sisters which I have labeled "Hannah & Julia". Without the ability to resize that left bar, it truncates the names with a `...` which is annoying. 

On macOS you can actually drag the left panel as wide as 500px. So I'd recommend either building that resize into the UI directly, or at the very least allowing the number to be changed in the settings file? 

image

PR #22, opened 2026-09-03
Pinned conversations survive shallow polls; first poll after start is deep

adamgamble:
## Problem

Pinned tiles popped in a beat after the panel opened. Pin metadata only comes from `imsg chats`, which the collector fetches on **deep** runs (a surface already open). Every shallow poll — the 6 s timer and each `imsg watch` push while the panel is closed — rebuilt threads from the message window with `pinned: false`, and BarWidget swapped its model for that list. So opening the panel rendered a dozen unpinned rows, then the open-triggered deep run re-pinned Shawn and Monica and grew the list to 300 a second later. (Live check: shallow run → 13 threads, none pinned; deep run → 300 threads, 2 pinned.)

## Fix

- **collector.ts** — `pins` (chat id → pin order, ids only, no content) is cached in `state.json` beside `groups` and `chatAliases`. Refreshed with the chat list on deep runs, re-applied to a shallow poll's rows by `applyPins()` and sorted through `compareThreads`. Output gains `deep: boolean` so the widget knows whether it received the complete list.
- **BarWidget.qml** — keeps the last deep list (`deepThreads`) and *overlays* a shallow result onto it by chat id instead of replacing it. A row the poll did not return has been read since the deep run (the ledger's catch-up fetch puts every unread chat inside the window), so its count is zeroed. The merged list sorts with a mirror of `compareThreads`.
- **BarWidget.qml** — the poll timer goes deep until one deep result has landed, so the first open after `omarchy-restart-shell` finds the complete, pinned list already in memory.

## Verified

- 25 s after a shell restart with the panel never opened: `status` reports `threads=300`, `state.json` holds the four pins.
- Screenshot 0.4 s after opening the panel shows all four pinned tiles with photos; unchanged at 3 s and 8 s. Plugin log grep clean.
- `bun test`: 287 pass (8 new for `pinsFromChats` / `applyPins` / `validPins`; existing `loadState` fixtures gain `pins`).

### DIFF
--- a/BarWidget.qml
+++ b/BarWidget.qml
@@ -207,6 +207,39 @@ BarWidget {
   // `readChat` clears one thread's blue dot — iMessage semantics: the dot
   // survives viewing the list and only goes when that conversation is opened.
   property var refreshQueue: []        // stateful requests are never overwritten
+  // The complete conversation list (quiet chats, pins) only comes from a deep
+  // run (`imsg chats`); a shallow poll returns the window's rows. Kept so a
+  // shallow result can be overlaid rather than replacing the model.
+  property var deepThreads: []
+  function overlayThreads(deep, shallow) {
+    var byChat = {}
+    for (var i = 0; i < shallow.length; i++) byChat[String(shallow[i].chat)] = shallow[i]
+    var out = []
+    for (var j = 0; j < deep.length; j++) {
+      var t = deep[j], c = String(t.chat)
+      if (byChat[c] !== undefined) { out.push(byChat[c]); delete byChat[c]; continue }
+      // Every unread conversation is inside the poll window (the ledger's
+      // catch-up fetch guarantees it), so a row the poll did not return has
+      // been read since the deep run.
+      out.push(t.unread > 0 ? Object.assign({}, t, { unread: 0 }) : t)
+    }
+    for (var k in byChat) out.push(byChat[k])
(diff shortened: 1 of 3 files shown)


nixfred:
Merged in 29f4e08 — thank you, and this one lands harder than the diff suggests.

I had seen this exact symptom yesterday: I opened the panel to check something, the pinned tiles were missing, and I wrote it off in my own notes as "the panel opened before the first deep poll finished." I filed it as a race and moved on. It was not a race, it was a design gap, and your description made that obvious in one read — pin metadata only arrives with `imsg chats`, deep runs only, and every shallow poll was replacing the model with a list that had `pinned: false` on every row.

Three things I want to note because they are the difference between a fix and a good fix:

- **Overlaying the shallow result onto the last deep list** rather than replacing it. Replacing was the original sin; a cache alone would have re-pinned the rows but still collapsed 300 threads to 13 on every poll.
- **Zeroing the count for a row the poll did not return**, with the reasoning spelled out (the ledger's catch-up fetch puts every unread chat inside the window, so an absent row has been read). That is the kind of thing that silently rots later if the argument is not written down.
- **Caching ids only, no content.** `state.json` never holding message content is a hard invariant here and it would have been easy to cache the whole chat row for convenience.

Merged by hand for one conflict in `collector.test.ts`, where we had both appended a describe block. 294 tests green.

That is four from you now — EXIF orientation, the bubble corner, Family Sharing contacts, and this.

PR #23, opened 2026-09-04
Rendering fixes: Retina-scale media, readable link colors, natural link artwork

jondkinney:
The three rendering fixes from #21, extracted as their own PR as requested, rebased onto current main (post-2.3.1).

**Retina media draws at its intended logical size.** A 144-DPI iPhone screenshot is a 2x image, but Qt sized it by decoded pixels, so it rendered as a near-full-width wall. `fetch.ts` now reads the image header at fetch time — PNG `pHYs` density, plus pixel dimensions for PNG/JPEG/GIF/WebP — and reports `pixelWidth/pixelHeight/pixelRatio` alongside the cached file (also on cache hits). The QML sizes attachments from those metrics instead of `implicitWidth`, which the `sourceSize` decode bound (the 800-unit texture cap) makes unreliable for exactly the large images this matters for. Falls back to `implicitWidth` when the header carries no metrics.

**Message links inherit the bubble's text color.** Qt's default theme blue was illegible on the accent fill; `richMessageHtml()` styles anchors with the bubble's own readable text color (the escaping/linkifying in `thread.ts` is unchanged — this only adds the style attribute).

**Tall link-preview artwork keeps its aspect ratio.** Portrait and square artwork was cropped to a shallow 220-unit banner; it now renders `PreserveAspectFit` with a deeper cap, so album art and portrait cards look like they do in Messages.

Two notes:

- The grouped-corner double-alpha fix from #21 is **omitted** — current main already solves it with per-corner radii (#10), so there is nothing left to fix.
- The first commit is the small ESM test-loader fix (also in #21) — without it `bun test` cannot load `linkpreview.ts` on main. With it, the full suite passes: 300 tests, 0 fail, including new unit tests for the PNG density parsing and QML assertions for all three fixes.

https://claude.ai/code/session_01R4wXLjBovV9ixsyJ4cqFVe

### DIFF
--- a/BlipView.qml
+++ b/BlipView.qml
@@ -94,6 +94,17 @@ FocusScope {
   readonly property color theirsFill: Qt.rgba(foreground.r, foreground.g, foreground.b, 0.14)
   readonly property color theirsText: foreground
 
+  // Links inside a bubble take the bubble's readable text color instead of
+  // Qt's default theme blue, which is illegible on the accent fill.
+  function richMessageHtml(html, linkColor) {
+    var color = String(linkColor || "").toLowerCase()
+    if (!/^#[0-9a-f]{6}$/.test(color)) color = "#ffffff"
+    return String(html || "").replace(
+      /<a href=/g,
+      '<a style="color: ' + color + '; text-decoration: underline;" href='
+    )
+  }
+
   readonly property string home: Quickshell.env("HOME")
   readonly property string threadScript:
     decodeURIComponent(Qt.resolvedUrl("thread.ts").toString().replace(/^file:\/\//, ""))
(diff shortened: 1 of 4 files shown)


nixfred:
Merged — thank you for splitting it, and for dropping the grouped-corner fix rather than carrying it: #10 already solved that on main and you noticed.

The Retina one is the find. Qt sizing by decoded pixels is the sort of thing you only catch by looking at a screenshot on a HiDPI panel and thinking "that is not the size it should be", and reading `pHYs` at fetch time rather than trying to guess in QML is the right layer. The note about `sourceSize` making `implicitWidth` untrustworthy for exactly the large images that matter is the detail that makes it correct rather than approximately correct — I would have used `implicitWidth` and shipped a subtler version of the same bug.

Link colour and aspect-preserved artwork both obviously right.

One correction for the record, since it is in your description: `bun test` on main is **not** broken without the ESM loader change — main runs 312 tests green using `require("./linkpreview")` inside the describe block. The import conversion is a fine cleanup and I took it, but it was not fixing a breakage. Possibly true against an older main, or against your branch before the rebase.

Rendering fixes done, so that is the first of the three. #24 next.

PR #24, opened 2026-09-04
Read-only contact review: cross-source duplicate scan, fingerprint cache, and card review

jondkinney:
Adds read-only contact review directly from a conversation: right-click its row or pin, or use the conversation header’s “Review contact” button. Group conversations let you choose a participant. Each matching card shows its source, account, matching-field count, and photo availability, with an exact-card shortcut to Contacts on the Mac.

“Scan contacts” checks conversation numbers and emails across Mac Contacts sources for possible duplicates and conflicting names. Named contacts and short codes participate in the scan. Results use the store-fingerprint and handle-set cache; message visibility is unaffected.

This revision removes the settings/preferences components and persisted display-name overrides. Configuration remains in `bridge.conf`. The bridge accepts only candidates, fingerprint, audit, and exact-card open operations; installation requires no Swift helper. Requests use bounded stdin and only bounded contact summaries enter the owner-only cache. The Mac contact scan, chat metadata fixes, and Python test coverage remain. The existing clipboard file-paste support is retained.

Validation: 382 Bun tests and 56 synthetic Python bridge tests pass, along with shell checks, Python compilation, JXA syntax, QML lint, and plugin validation. The actual review and scan screens were exercised and inspected using an isolated fake bridge. CI now runs every Python bridge suite. Rebased onto main at `438f5c0`.

Contact editing, comparison, and cleanup remain in draft #25.

### Menubar layout

Contact review uses compact, wrapping action buttons and a separate scan footer. Its panel height stays independent of the message list. Verified with a synthetic long-name contact at the normal panel width; 417 Bun tests pass.


nixfred:
Update on this one, and the news is good: the hold is lifted and I want the contact review.

I verified the read-only claim properly rather than taking it on trust, and it holds up. `contacts` exposes only `list find lookup resolve dump sources`; the `resolve` stdin protocol accepts `candidates`, `fingerprint`, `audit` and `open`, all of which read; the single call into `contact-repair.js` is `{"operation": "available"}`, an availability probe, not a write. And you dropped the `swiftc` dependency from `install.sh` — thank you, that was the thing I was most worried about, since it would have made a working Xcode a requirement for anyone installing the bridge.

What held it up was the other half. This PR also carries the settings surface — `BlipSettings.qml`, `BlipPreferences.qml`, and `IdentitySettings.qml` at 1411 lines — and that was an open product question I had not answered yet. I have now, in #21, and it went against adding a settings system: Blip configures through `bridge.conf` keys, and that is written into CLAUDE.md as an invariant. Please read the #21 comment for the full reasoning, including the `hideShortCodeConversations` default that decided it. It is a real explanation, not a brush-off.

So the ask, and I am aware this is the third revision I have asked of you:

- **Keep:** the cross-source duplicate scan, the fingerprint cache, `contacts_test.py`, the `imsg` changes and `imsg_test.py`. That is the value here and I have nothing like it.
- **Drop:** `BlipSettings.qml`, `BlipPreferences.qml`, `preferences`-style persisted config.
- **Your call:** `IdentitySettings.qml` and `BlipIdentities.qml`. A contact-review feature needs somewhere to live and I am not pretending otherwise — if the review UI is reachable from the conversation surface rather than from a settings page, it is in scope. Show me the smallest version that works and I will judge it on that.
- **Rebase needed regardless:** main moved nine commits under you, including a compose rewrite (#28) and the font scaling (#30). Fair warning from my own bruises — those two collided badly enough that I shipped a merge conflict marker into `BlipView.qml` and blanked the whole panel for a release. If `BlipView.qml` fights you, that region is why.

If you would rather hand off the rebase than do it a third time, say so and I will do it and keep it under your name. Your call either way, and thank you for staying with this.

nixfred:
Picking this up again after clearing the queue — six PRs landed today, so main has moved a long way under this branch. Before anything else, one factual correction, because it changes what needs doing.

The description says "Rebased onto main at `438f5c0`". That is not what the branch is. Its merge-base with main is `bc3a498` — **Blip 2.3.2** — and it is a single squashed commit on top of that. I checked because I wanted to merge it:

```
$ git merge-base pr/24 main
bc3a498  Blip 2.3.2 — mute the blasts, and your own address book decides who people are
```

It also conflicted against `438f5c0` itself, so this was not something today's merges caused. Against current main it is **70 conflict hunks across 14 files, 50 of them in `BlipView.qml`**.

That is precisely the collision I warned about in my last comment — the compose rewrite (#28) and the font scaling (#30) are in that region, and today's optimistic-send work (#43) and keyboard cursor (#39) are now in there too. Fifty hunks in that file is not something I will resolve on your behalf: the failure mode is silent, and I would be guessing at your intent in a 2822-line feature I did not write. I have just spent a merge doing exactly that kind of integration for #39 and it turned up two breaks that no test caught until I went looking.

So the ask is a real rebase onto current main, not another squash onto 2.3.2. Everything else in my previous comment still stands and still holds — the read-only claim, the dropped `swiftc` dependency, the scan and the fingerprint cache are all things I want.

Two things I noticed reading the current head that are worth your attention while you are in there:

1. `audit-cache.json` is the first contact data Blip writes to disk. It is 0600 and bounded and it is not message content, so it does not cross the invariant — but it is a new class of data at rest and it should be named as such in `docs/PRIVACY.md` rather than only in the code comments.
2. `contact-repair.js` reaches Contacts through JXA. First run will fire an Automation prompt at whoever installs it, and per the `#36` note in `CLAUDE.md`, a prompt nobody answers in ~2 minutes is recorded by TCC as a *denial* that the Settings switch may then refuse to undo. Worth a line in the setup path telling people to be at the Mac, the way `blip-setup` does for the Messages grant.

My offer from last time stands: if you would rather hand the rebase off than do it a third time, say so and I will do it and keep it under your name.

nixfred:
Following up so this isn't sitting silently — the read-only framing here is right, and I'd like to take it.

It needs a rebase before I can. `main` has moved 26 commits since you branched, and the merge stops on four files:

    .github/workflows/ci.yml
    CHANGELOG.md
    bridge/mac/imsg
    collector.ts

`CHANGELOG.md` and the workflow file should be mechanical. `bridge/mac/imsg` and `collector.ts` are the two that moved under you and are worth a real look rather than taking either side wholesale.

No hurry, and if you'd rather I did the rebase myself say so and I will — I held off because the duplicate-scan and fingerprint-cache logic is your design and I didn't want to resolve conflicts by guessing at your intent.

Leaving #25 alone while it's a draft; it stacks on this one anyway.

nixfred:

PR #25, opened 2026-09-04
Gated contact writes: edit, merge, clean up, delete, and undo

jondkinney:
Adds an optional contact workspace for editing source cards, consolidation, deletion, narrow phone/email cleanup, Apple's Link/Merge handoff, and undo. Open it through a conversation’s Contact review → Manage contact. Person selection lasts only for the current session.

The settings/preferences system and saved display-name overrides have been removed from this branch too. The existing comparison/editor and native mutation contracts remain: bounded stdin requests, exact card tokens and revisions, preview/apply confirmation, independent Linux and Mac write gates, and private Mac undo receipts. Near-duplicate address consolidation and its behavior tests are preserved.

Messaging and #24’s read-only scan need no Swift helper. This draft’s full card comparison, vCard export, and native mutations use the optional compiled Contacts helper; installation skips compilation when Swift is unavailable. Contact writes require explicit setup opt-in. Delete/consolidation undo recreates text fields but cannot restore original accounts, links, or photos; Apple's UI Link/Merge handoff has no automatic Blip undo.

Validation: 413 Bun tests and 84 synthetic Python bridge tests pass. Python compilation, shell syntax/ShellCheck, JXA syntax, and plugin validation pass. Review, scan, management, and comparison screens were exercised and inspected with an isolated fake bridge; no real contacts or conversations were accessed.

Remains **draft** pending the read-only extractions below and Mac integration validation of mutation failure/recovery behavior. #24 has merged. The Mac write paths were not exercised during this revision.

### Read-only functionality extracted for separate review

- #59 removes the 200-handle total scan limit by batching bounded Mac requests.
- #60 adds full read-only source-card details without the optional Swift helper.
- #61 restores Copy vCard for an explicitly selected source card without enabling contact writes.

This draft still contains its earlier detail/export implementations. Once #60 and #61 land, rebase this branch onto main, reuse those read-only paths, and remove the duplicated viewers/export plumbing. The remaining diff should focus on editing, merge/consolidation, cleanup, delete, undo, and their write gates. The write-path validation below describes the existing draft; no new Mac mutation validation has been performed.

### Contact management layout and navigation

Quick review stays in the menubar. “Manage contact ↗” opens that contact directly in the detached Blip window. An existing workspace retains its editor or preview when another contact is requested. Startup handoffs queue behind the initial access check.

One matching person opens directly to their source cards; conflicting matches still require a choice. Single cards show contact details without comparison controls. Action rows stack or wrap, and two-card comparisons adapt to the window width. Shared buttons support keyboard focus.

Validated with synthetic screenshots of review, single-card details, comparisons at 720 and 1040 window widths, and the editor. 452 Bun tests and 92 Python tests pass, including window handoff lifecycle coverage. Preview, exact-token validation, local/Mac write gates, and explicit confirmation remain in place.

### DIFF
--- a/BarWidget.qml
+++ b/BarWidget.qml
@@ -210,6 +210,10 @@ BarWidget {
       'for i in 1 2 3 4 5 6 7 8; do a=$(hyprctl clients -j | jq -r \'.[] | select(.title | startswith("Blip")) | .address\' | head -1); [ -n "$a" ] && break; sleep 0.15; done; ' +
       '[ -n "$a" ] && hyprctl dispatch "hl.dsp.focus({ window = \\"address:$a\\" })" >/dev/null'])
   }
+  function manageContact(handle) {
+    showApp()
+    return windowLoader.item !== null && windowLoader.item.manageContact(handle)
+  }
   /** Either surface open → keep the deep (complete) thread list. */
   function anySurfaceOpen() {
     return (panelLoader.item && panelLoader.item.opened === true)
--- a/BlipView.qml
+++ b/BlipView.qml
@@ -1755,7 +1755,7 @@ FocusScope {
   // press past the last row landing at the top reads as a jump, not a loop
   // (Omarchy's Dropdown clamps the same way).
   function moveCursor(dy) {
-    if (contactReview.opened || !listShowing || threads.length === 0 || dy === 0) return
+    if (contactsOpen || !listShowing || threads.length === 0 || dy === 0) return
     // Up from the first row hands focus to the search field above the list,
     // and Down in an empty field hands it back (Omarchy's SearchableDropdown).
     if (dy < 0 && cursor <= 0) { startSearch(); return }
@@ -3421,6 +3421,13 @@ FocusScope {
     }
   }
 
+  // Contact work owns the full detached surface. Refuse replacing an open
+  // workspace so another menubar click cannot discard an editor or preview.
+  function openContactManagement(handle) {
+    if (!root.splitView) return false
+    return contactWorkspace.review(handle)
+  }
+
   property var contactContext: null
   Menu {
     id: contactMenu
@@ -3440,7 +3447,12 @@ FocusScope {
     fontFamily: root.fontFamily
     fontSize: root.fontBodySmall
     onClosed: root.focusDefault()
-    onManageRequested: function(handle) { contactWorkspace.review(handle) }
+    detached: root.splitView
+    onManageRequested: function(handle) {
+      if (root.splitView) root.openContactManagement(handle)
+      else if (!root.hostWidget || !root.hostWidget.manageContact(handle))
+        contactReview.notice = "Finish the contact already open in the Blip window first."
+    }
   }
   ContactWorkspace {
     id: contactWorkspace
--- a/BlipView.qml
+++ b/BlipView.qml
@@ -84,7 +84,7 @@ FocusScope {
   readonly property color dim: Qt.darker(foreground, 1.45)
   /** An editor owns the keyboard — the host's key catcher must stand down. */
   readonly property bool editorActive:
-    contactReview.opened || composeField.activeFocus || searchField.activeFocus || newField.activeFocus || bubbleFocused
+    contactWorkspace.opened || contactReview.opened || composeField.activeFocus || searchField.activeFocus || newField.activeFocus || bubbleFocused
   readonly property alias composeEditor: composeField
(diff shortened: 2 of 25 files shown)


nixfred:
Holding note so you do not spend a rebase on this prematurely — main has moved nine commits and this PR is +14686, so that would be a real cost for nothing.

This one queues behind #24. The read-only review is the piece I have said yes to in principle; a write path into my actual address book is a separate decision and it is mine to make deliberately, not as a follow-on. Two things I want to see settled before I judge it:

1. What #24 looks like once the settings layer comes out and it is rebased. If the review surface ends up small and readable, the case for writes gets easier to evaluate on its own terms.
2. The failure modes. Editing and merging contact cards is the one operation here that can lose data that is not recoverable from a Mac backup, and I want to be able to answer "what happens when this goes wrong halfway" before it merges, not after.

None of that is a no. It is that #24 and #25 in flight together means judging a write path against a moving read path.

Also relevant: the settings surface question is now closed — Blip configures through `bridge.conf` keys, no preferences file, no settings panel. Reasoning is in #21. If any part of this PR routes through that surface, it will need the same treatment #24 is getting.

Genuinely appreciate how much work is sitting in this branch, and I am sorry it is waiting. It is waiting on my decision, not on its quality.

nixfred:

jondkinney:
compare-narrow
compare-wide
contact-details
contact-editor

nixfred:
Read this properly rather than letting it keep sitting. Here is where I land, and why.

**The delete path is the one thing I will not take.** Everything else here I want.

Your verification is not the problem. `contact-delete.swift` runs `store.execute(save)` and then `requireMissing(identifiers, in: store)` at `:286`, refetching every identifier and throwing code 19 if any survive, and the consolidate path additionally proves the survivor exists with `exactContact(targetUid…)`. Refetching sources in a fresh `CNContactStore` session per attempt, with the comment about the Cocoa 134092 stale-relationship failure, tells me you hit the real failure modes and handled them. That is careful work.

The problem is the undo, measured against what delete actually does. `CNSaveRequest.delete()` is permanent: Contacts has no Recently Deleted, and it propagates over iCloud to every device signed into that account. Against that, what Blip offers is:

- `UNDO_FILE = "latest.json"` (`bridge/mac/contacts:73`) is a **single slot**. All four write paths call `write_undo_receipt`, which `os.replace`s over it. Delete a card, do any second contact write, and that receipt is gone.
- Undo for a delete does not restore. `undo_contact_removal` calls `run_contact_repair({"operation": "restore", …})`, which runs `createPerson` and builds a **new** card from captured fields. New identifier, container chosen by Contacts, and as your own PR body concedes, no original accounts, links or photos.
- The Apple Link/Merge handoff has no Blip undo at all.

So: one level of undo, lossy, and only until the next write. That is not a safety net for a permanent cross-device deletion, and I am not shipping it as one. Edit, merge and cleanup are a different risk entirely — their worst case is a mangled card, and a re-created card is a genuine repair for that.

**What I would take, and I think it is most of your work:**

1. Split delete and the Apple Link handoff out of this PR. That also drops `contact-delete.swift`, which means dropping the `xcrun swiftc` build step from `install.sh`, and I would rather not ship a compiled helper for a path I am declining anyway.
2. Rebase onto main. Your own precondition has been met: **#59, #60 and #61 all merged.** Your description said to rebase once #60 and #61 landed, reuse those read-only paths, and remove the duplicated viewers and export plumbing, leaving editing, merge, cleanup and their write gates. That is exactly the PR I want. The branch is 127 commits behind and a meaningful share of these 8,868 lines is code that now exists twice.
3. Keep `require_write_gate()` and `~/.blip/contact-writes-enabled`. Making the whole surface off unless explicitly armed is the right design and I want it kept.

Two things to fix in the reduced PR:

- The receipts write contact PII (names, phones, emails) in plaintext to `~/.blip/contact-undo`. The file handling itself is careful, 0700 directory, `O_NOFOLLOW`, `dir_fd`, `fsync`, uid checks, atomic replace. But it is a new on-disk PII store outside Contacts with a TTL and no pruning on read. Say so in `docs/PRIVACY.md`.
- If receipts survive at all, give them a history rather than one slot, even for edit and merge. Losing the receipt for a mangled card on the next write is the same bug, just with a recoverable blast radius.

I know this is a large PR to be told to cut down, and I would rather say it plainly than leave it open for another month. Your read-only work is already in and being used. The write gates and the verification discipline here are good enough that I want the rest of it landed, just not the one operation that cannot be taken back.

PR #27, opened 2026-09-04
Mute list: hide spam conversations (ActBlue / "Stop2End" blasts) entirely

tlehman:
## What this is for

Political fundraising spam. The texts that read *"the deadline is TONIGHT, rush \$25 — Reply STOP2END"*, arriving from a five-digit short code you have never seen, which will be a **different** short code next week. Nobody opted into them, `allowlist.json` cannot help (it gates toasts, not the sidebar), and blocking the number is whack-a-mole because the number is disposable.

The two things that are *not* disposable are the platform's name (`ActBlue`, `WinRed`) and the opt-out footer the TCPA makes every one of these carry (`Stop2End`, `Reply STOP`). So the mute list matches **phrases** as well as handles, and that is what makes it work.

## What it does

`~/.config/blip/mutelist.json` — the allowlist's mirror image in shape, its opposite in effect:

```jsonc
// re-read every poll, no restart
{ "mute": ["ActBlue", "WinRed", "Stop2End", "78462"] }
```

Same parser shape as `allowlist.json` (bare array or `{ "mute": [...] }`, non-strings dropped, a missing file is an empty list, re-read every poll). Where the allowlist says *who may raise a toast*, the mute list says *whose conversation Blip shows at all*: a matched chat has no sidebar row, no unread count, no toast.

**Matching**, and the reasoning behind each rule:

- a handle or chat id matches **exactly**, as in the allowlist — a substring of a phone number can never take out an unrelated number;
- a phrase matches **case-insensitively anywhere in the text**, two characters minimum, so a stray one-character entry cannot mute the world;
- **inbound only** — forwarding *"another ActBlue text, unbelievable"* to a friend must not mute the friend;
- **one match mutes the whole conversation**: a blast carries its footer on some messages and not others, and half a thread left in the sidebar is worse than none;
- a chat row whose newest message is **your own reply** survives.

## Where the cut happens

In `collect()`, on the deduped window **before** `unreadCounts`, `buildThreads` and `selectToasts` — and on the `imsg chats` rows a deep run merges in. That second half matters: the chat list reaches back further than the message window, so without it a blast last seen a month ago would reappear the moment the panel opened. Filtering upstream of the ledger means there is no second filter in the QML to keep in sync, and a muted conversation cannot badge the bar.

The watermark still advances across muted rows, so unmuting later does not replay them as a pile of new toasts.

## Invariants

- **No message content on disk.** `mutelist.json` holds only what the user typed; `state.json` is untouched. Added to the `docs/PRIVACY.md` table.
- **Logic in TypeScript, QML only renders.** No QML change at all — the widget renders the thread list it is given.
- **Nothing is deleted.** The messages are untouched on the Mac and still in Messages; Blip stops showing them.

## Tests

14 new tests in `collector.test.ts`, in the allowlist tests' style: loader shapes, exact-vs-phrase matching, the one-character guard, outbound quoting, whole-conversation muting, chat-row filtering by id / alias / preview text, the own-reply survivor, `null` on a shallow poll, and the end-to-end "a muted blast never reaches `selectToasts`". `bun test`: **309 pass, 0 fail**.

Docs: README section next to the toasts one, a `CHANGELOG.md` Unreleased entry, and the privacy table row.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,20 @@
 
 ## Unreleased
 
+- **A mute list, for the texts nobody opted into.** Political fundraising
+  blasts — "the deadline is TONIGHT, rush $25, Reply STOP2END" — arrive from a
+  short code that is different every cycle, so blocking the number is
+  whack-a-mole. `~/.config/blip/mutelist.json` is the allowlist's mirror image
+  in shape (`{ "mute": [...] }` or a bare array, re-read every poll, no
+  restart) and its opposite in effect: a matched conversation is dropped
+  before the unread ledger, the thread list and the toasts are built, so it
+  cannot show, count, or interrupt. Entries match a handle or chat id exactly,
+  or a phrase of two or more characters anywhere in an INBOUND message
+  (`"ActBlue"`, `"WinRed"`, `"Stop2End"`) — the platform's name and the
+  legally required opt-out footer both outlive the rotating number. One match
+  mutes the whole conversation; quoting a muted phrase to a friend never mutes
+  the friend. Nothing is deleted: the messages stay untouched on the Mac.
+
 - **Your own address book decides who someone is.** Contact names were
   resolved by counting sources and taking the most common spelling, so a work
   Exchange account or a synced company directory holding thousands of rows
--- a/README.md
+++ b/README.md
@@ -353,6 +353,28 @@ else still counts and still shows.
 { "allow": ["+15551234567", "them@icloud.com"] }
 ```
 
+**Mute (spam)** — the allowlist's opposite: a muted conversation does not
+show at all. No sidebar row, no unread count, no toast. This is the knob for
+political fundraising blasts — the "the deadline is TONIGHT, rush $25,
+Reply STOP2END" texts that arrive from a short code you have never seen and
+that will be a different short code next week. Listing the number is
+whack-a-mole, so list the words instead: the PAC platform's name (`ActBlue`,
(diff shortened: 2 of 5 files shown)


nixfred:
Merged in 16d6831 — thank you. This is a well-argued PR and the argument is the valuable part.

The insight that carries it: **the short code rotates, the opt-out footer does not.** Matching on `Stop2End` rather than on a number is the difference between a feature and whack-a-mole, and I would not have got there on my own.

Two implementation choices I want to name because they are the ones that would have been easy to get wrong:

- **Cutting upstream of the ledger**, before `unreadCounts` / `buildThreads` / `selectToasts`, rather than filtering in the QML. It means there is no second filter to keep in sync and a muted blast physically cannot badge the bar.
- **Advancing the watermark across muted rows anyway**, so unmuting later does not replay a month of blasts as a pile of toasts. That is the detail I would have missed, and it is the kind of thing you only think of if you have actually been bitten.

Also right: filtering the `imsg chats` rows separately, since the chat list reaches back further than the message window.

I added one thing on top (a follow-up commit, not a change to yours). Your docs cover the case where *you* quote a muted phrase outbound — correctly, that never mutes the person you said it to. The mirror case is not written down: matching is on inbound text, so a *friend* writing "I got another ActBlue text today" mutes your friend's whole conversation until you edit the list. That is exactly what phrase matching means, but someone hitting it would read it as a bug, so the README now says to pick phrases nobody would type at you, and there are three tests locking the behaviour so it stays a decision.

Shipping in 2.3.2 tonight.

PR #28, opened 2026-09-04
compose: wrap long text and attach copied files

jefehoser:
## What

Two Linux compose-box fixes, both found while using Blip on Omarchy.

1. **Copied files attach instead of pasting a path.** File managers put `text/uri-list` (and often a `file://` path as `text/plain`) on the clipboard, not `image/png`. Ctrl+V was inserting that path as a text message. The clipboard helper now treats a local file URI/path as an attachment, the same as drag-and-drop.
2. **Compose wraps at the box edge.** The field was a single-line `TextField`, so long text scrolled sideways. It is now a width-capped wrapping `TextArea`. Enter still sends; Shift+Enter inserts a newline.

## Why

Pasting a photo from the file manager sent the local path as iMessage text. Typing a normal sentence ran off the right edge of the compose box.

## How it was verified

- [x] `bun test` is green (CI will check) — 315 pass, including new paste/uri-list tests and a QML wrap invariant
- [ ] UI change → screenshot attached — verified live on Omarchy (file paste + wrap); no screenshot of a real conversation

No `CLAUDE.md` invariant change. Message text still stays off argv; compose remains never-disabled (`readOnly` while sending).

### DIFF
--- a/BlipView.qml
+++ b/BlipView.qml
@@ -1111,7 +1111,7 @@ FocusScope {
         if (!root.inThread || String(root.active.chat) !== root.pasteChat) return
         try {
           var d = JSON.parse(text.trim())
-          if (d.kind === "image") root.setDraft(String(d.path || ""))
+          if (d.kind === "image" || d.kind === "file") root.setDraft(String(d.path || ""))
           else if (d.kind === "text") {
             composeField.insert(composeField.cursorPosition, String(d.text || ""))
           }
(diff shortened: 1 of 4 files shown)


nixfred:
Merged in f067226 — thank you. Sending the file **path** as an iMessage is the kind of bug that is obvious the moment someone hits it and invisible until then; `text/uri-list` being what a file manager actually puts on the clipboard is the detail that makes the fix right rather than a special case. The single-line TextField was mine and I had stopped seeing it. Merged by hand for a one-line import conflict: #23 had added `imageMetrics` to the same import an hour before you. 321 tests.

PR #29, opened 2026-09-04
avatars: find group photos and retry after a miss

jefehoser:
## What

Group tiles with a real Messages picture show that picture instead of a letter. A photo assigned after Blip first asked (a new group picture or a Contacts card) shows up on the next open instead of staying a letter for a day.

## Why

Two stacked misses:

1. Messages stores a group's photo on an announcement row (`item_type = 3`). The bridge hides those rows so they never become empty unread bubbles, and the photo lookup joined through that same filter, so it never saw `GroupPhotoImage`. A multi-megabyte PNG also blew the 2MB Linux avatar cap.
2. The first "no photo" was cached for a day on disk and for the rest of the session in the window, so a picture set a minute later was ignored.

Groups still never borrow a member's face. Collages that Messages draws from members are not a stored image and still show initials.

## How it was verified

- [x] `bun test` is green (CI will check)
- [x] Mac-side lookup unit test (`python3 bridge/mac/test_group_photo.py`) — no live `chat.db`
- [x] Live `imsg avatar --chat` / Contacts fetch against already-assigned photos (metadata and image magic only; no conversation contents)

Invariants: group avatars are the group's own photo, never the last speaker; `MESSAGE_SRC` still hides announcements from the transcript.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,6 +15,8 @@ jobs:
         run: bun test
       - name: Mac tools compile (python3 syntax, no macOS needed)
         run: python3 -m py_compile bridge/mac/imsg bridge/mac/imsg-send bridge/mac/contacts bridge/mac/tcc-check bridge/mac/blip-check bridge/mac/blip-dispatch
+      - name: group-photo lookup (no macOS needed)
+        run: python3 bridge/mac/test_group_photo.py
       - name: shell scripts (bash -n + shellcheck)
         run: |
           bash -n bridge/linux/blip-shim scripts/blip-setup scripts/sync-bridge.sh bridge/mac/install.sh
--- a/BlipView.qml
+++ b/BlipView.qml
@@ -473,10 +473,25 @@ FocusScope {
     avatarQueue.push(handle)
     pumpAvatar()
   }
+  // Letters stick in avatarFiles as "". Opening the panel/window again drops
+  // those and re-asks, so a photo set a minute ago is not stuck until tomorrow.
+  function retryBareAvatars() {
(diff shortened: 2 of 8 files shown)


nixfred:
Merged in f69cf34 — thank you, and this one is a regression I shipped, so I owe you the detail.

I landed the `GroupPhotoImage` lookup on 09-02, then merged #14 (hide announcement rows — they were counting as unread forever; 811 of them on my Mac) on 09-03. Both were right on their own. Together they were not: the photo lives on an `item_type = 3` announcement row, and my lookup joins through the same `MESSAGE_SRC` that #14 taught to hide them. I never checked one against the other, and group photos have been letters since 2.3.0 shipped.

Verified before and after on my own Mac:

```
LMT via main         →     0 bytes
LMT via this PR      → 33881 bytes
Sportsball! via this → 33023 bytes
```

Both are back on my screen now. Note the size — your resize also fixes the 2 MB Linux avatar cap that a multi-megabyte group PNG was blowing, which is the second miss in your description and one I would not have connected to the first.

The retry change is the part I like most as a design decision: caching "no photo" for a day made sense for a Contacts card that genuinely has no picture, and made no sense at all for a group picture someone sets a minute later. Distinguishing those two is the right call rather than just shortening the TTL.

Also appreciated: `bridge/mac/test_group_photo.py` runs without a live `chat.db`. There has been no Python test harness in this repo at all, and after three Mac-side bugs in two days that gap has been on my mind. Yours is the first.

Deployed to my gateway. Shipping in the next release.

jefehoser:
Thanks for taking all three, and for the notes — especially this write-up. Glad they helped. I really like what you’re doing with Blip!

PR #30, opened 2026-09-04
ui: optional Blip-only type size in bridge.conf

jefehoser:
## What

`ui_font_size=14` in `bridge.conf` sets Blip bubble text in pixels (9–24). Names and timestamps scale with it. Unset, Blip still follows Omarchy's type tokens. The bar, GTK apps, and terminals are unchanged.

## Why

Blip was locked to Omarchy `bodySmall` (~11px). Signal's message text is larger, and `omarchy display text size` would scale the whole desktop. This is a Blip-only knob, next to the existing `ui_font=theme` family override.

## How it was verified

- [x] `bun test` is green (CI will check)
- [x] Live window at `ui_font_size=14` after `omarchy restart shell`

No send/receive behavior. No Mac-side tools.

### DIFF
--- a/BarWidget.qml
+++ b/BarWidget.qml
@@ -632,6 +632,8 @@ BarWidget {
   property bool automationOn: false
   /** `ui_font=theme` in bridge.conf: never use SF Pro even when it is present. */
   property bool uiFontTheme: false
+  /** `ui_font_size=N` in bridge.conf: bubble text in px (9–24). 0 = Omarchy default. */
+  property int uiFontSize: 0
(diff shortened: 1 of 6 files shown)


nixfred:
Merged in 03bb3e1 — thank you. Two things I appreciated: the default is genuinely unchanged, so this costs nothing for anyone who does not set it, and you put it next to `ui_font=theme` in `bridge.conf` rather than inventing a second config surface. `omarchy display text size` really is the wrong hammer for "Blip's messages are small". The 9–24 clamp being enforced rather than documented is the right call too. Merged by hand for a one-hunk CHANGELOG conflict after #28 and #29 landed the same hour.

PR #31, opened 2026-09-04
Contacts saved without a country code get their name and photo

Fileri:
## What

Contacts saved without a country code now get their name and photo. `imsg` keeps matching on the exact last-ten key first; when that finds nothing, it matches the way Contacts does: the Mac's own region fills in the missing country code, and two numbers are the same when their calling codes agree and one national number ends with the other, a saved trunk zero (`07700 900123`) included.

Nothing that resolves today changes — the exact key still decides whenever it hits.

## Why

The last-ten key is a whole national number in North America and nothing anywhere else. A card saved as `123 45 678` never met the handle `+4712345678`, so that person showed as a bare number with initials — the owner's own card included, which is how this was found. On one Norwegian address book, 99 of 386 numbers are saved without a country code.

## How

- **`bridge/mac/calling_codes.py`** (new) — region → calling code for 245 regions, generated from libphonenumber v9.0.38; the regeneration one-liner is in its header. The set is prefix-free, so splitting a `+` number into calling code and national number is unambiguous. `imsg` imports it as a sibling; `install.sh` and the CI compile step know about it.
- **`bridge/mac/imsg`** — `_parse_phone()` splits `+…`/`00…` on the calling code and gives a bare number the Mac's region (`AppleLocale`, `rg=` override honoured). `_same_number()` is libphonenumber's SHORT_NSN_MATCH with two guards: a floor of seven shared digits, and a North American card shorter than ten digits is refused rather than guessed (it is missing its area code; Contacts would guess). `name_for` keeps a second index of every saved number, parsed, for the fallback. `cmd_avatar`'s candidate selection is now `_avatar_candidates()` and follows the same rule, exact first per source.
- **`bridge/mac/test_phone_match.py`** (new) + one CI step, same shape as `test_group_photo.py`.
- **Not touched**: the `contacts` CLI. Blip never calls its lookup.
- README (Outside North America), CLAUDE.md (new invariant), CHANGELOG (Unreleased).

## How it was verified

- [x] `bun test` is green (CI will check) — 323 pass; `test_phone_match.py` 19 pass; `test_group_photo.py` pass; every Mac tool compiles
- [x] Any send/receive behavior was exercised against **my own number only** — this change does not send; it only resolves names and photos
- [ ] UI change → none; the same tiles and rows render better data
- [x] Touches an invariant in `CLAUDE.md` → **new: phone numbers match the way Contacts matches them**; the cross-source ambiguity rule is unchanged

Against a live `chat.db`, old tool vs new, through the real `name_for` on every distinct phone handle (1393): **1356 resolve as before, 37 gain a name, 0 lose one, 0 change to a different name.** In the conversation list, unnamed phone threads went 9 → 6; the three that gained are people whose cards are saved without a country code. The owner's own photo now streams (0 → 58 KB); a card saved with `+47` returns byte-identical bytes before and after. A Dutch-looking handle ending in the same eight digits resolves to nothing.

Cross-checked against Contacts itself on macOS 26, via `CNContact.predicateForContactsMatchingPhoneNumber` on the owner's own card (saved without a country code): the `+47` handle and the national form both resolve to the card; the same-length number with a leading digit changed, the same-length number with two digits changed, and a US number ending in the same ten digits all resolve to nothing. The rule in this PR gives the same five answers.

The 19 tests cover: the table is prefix-free and splits correctly (incl. Iceland's 7-digit national numbers); region from `AppleLocale` in five spellings and a missing prefs file; a US number in four spellings; the region rule; `00` prefixes; exact beats suffix; two matching cards name nobody; a same-length different number never matches; another country never matches; the UK trunk zero; the Italian zero; the North American refusal; the no-region fallback; and the photo path three ways. Example numbers are unassignable (`+47 1xx…`), the UK drama range, or 555-01xx.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,9 +14,11 @@ jobs:
       - name: unit tests (collector, thread, tier2, ui invariants)
         run: bun test
       - name: Mac tools compile (python3 syntax, no macOS needed)
-        run: python3 -m py_compile bridge/mac/imsg bridge/mac/imsg-send bridge/mac/contacts bridge/mac/tcc-check bridge/mac/blip-check bridge/mac/blip-dispatch
+        run: python3 -m py_compile bridge/mac/imsg bridge/mac/imsg-send bridge/mac/contacts bridge/mac/tcc-check bridge/mac/blip-check bridge/mac/blip-dispatch bridge/mac/calling_codes.py
       - name: group-photo lookup (no macOS needed)
         run: python3 bridge/mac/test_group_photo.py
+      - name: phone matching, the way Contacts does it (no macOS needed)
+        run: python3 bridge/mac/test_phone_match.py
       - name: shell scripts (bash -n + shellcheck)
         run: |
           bash -n bridge/linux/blip-shim scripts/blip-setup scripts/sync-bridge.sh bridge/mac/install.sh
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,19 @@
 
 ## Unreleased
 
+- **Contacts saved without a country code get their name and photo.** The
+  bridge matched phone numbers on their last ten digits, which is a whole
+  national number in North America and nothing anywhere else: a card saved
+  as `123 45 678` never met the handle `+4712345678`, so that person — the
+  owner's own card included — showed as a bare number with initials. When the
+  exact key finds nothing, `imsg` now matches the way Contacts does: the Mac's
+  region fills in the missing country code (`calling_codes.py`, generated from
+  libphonenumber), and two numbers are the same when the calling codes agree
+  and one national number ends with the other — a saved trunk zero
+  (`07700 900123`) included. A North American card shorter than ten digits is
+  missing its area code and is refused rather than guessed. The exact key
+  still decides first, so nothing that resolved before changes. Mac side:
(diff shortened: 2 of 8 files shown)


nixfred:
Merged as 476a839 — thank you. This is the most careful of the four and the one I read slowest, because name resolution is where a wrong match does real damage: a stranger's name on a conversation is worse than a bare number. What convinced me, in order: exact key first so nothing that resolves today can change; the seven-digit floor; the North American refusal (Contacts would guess, you chose not to — right call); and the cross-check against `CNContact.predicateForContactsMatchingPhoneNumber` giving the same five answers. That last one is the difference between "matches libphonenumber" and "matches what the user sees in Contacts."

Verified here: `test_phone_match.py` 19/19 on Linux, the prefix-free check on the calling-code table, and against my live Mac: 0 names changed for anything that was named before, 0 gained (my region is US and my unnamed handles are short codes and services, so the guard correctly refuses all of them), 299 threads before and after. `calling_codes.py` is on my Mac beside `imsg` and both compile there. The `install.sh` addition was the one thing I checked twice — a fresh install without the sibling module would import-error on every call — and it is right. Merged by hand (CHANGELOG, the CI step list, and `import functools` beside `import html` from #34), no rebase needed.

All four of yours today measured the problem on a real database before touching anything and said exactly what changed and what did not. That is the standard.

PR #32, opened 2026-09-04
Unsent messages are tombstones again, not empty bubbles

Fileri:
## What

Unsent messages render as the tombstone Blip already has ("They unsent a message") instead of an empty bubble tagged "Edited". Nothing is hidden: the row stays in the thread, shown the way Messages shows it.

## Why

On macOS 26 an Undo Send no longer sets `date_retracted`. Apple records it as an edit: `date_edited` is stamped, `text` and `attributedBody` are cleared, and the withdrawn parts are listed as `rp` in `message_summary_info`. The bridge read only the two date columns, so it saw "edited, empty body" and passed a bubble through; the tombstone path in `BlipView.qml` never got its flag.

An unsent message in a conversation: a dim italic "Jamie Rivera unsent a message" line where the bubble was, with the run's timestamp under it

*Rendered by `scripts/demo/blip-shots` against the fake bridge — invented people, real code paths.*

On one macOS 26 database: `date_retracted` is zero on all ~33,600 rows; 14 rows carry `rp` (all since 2026-01) and no body; 8 rows carry `ec` (a real edit's history) and keep their body. Nothing carries either. Found as "a group with a few empty messages".

## How

- **`bridge/mac/imsg`** — `base_select` reads `m.message_summary_info` beside the two dates (`NULL AS …` on Macs too old to have it, the pattern the dates use). `_unsent(blob)` is true when the plist lists withdrawn parts in `rp`, false for anything missing or unreadable. `_edit_flags(date_edited, date_retracted, summary)` decides: retracted when either the old column or `rp` says so, else edited when `date_edited` says so — an unsend is not an edit, whatever `date_edited` says. The old `date_retracted` path is kept for older Macs. `enrich()` calls it; nothing else changes.
- **`collector.ts`** — the self-thread. Every message there lands twice, and `dedupeSelfEcho` folds the empty outbound row into its text-bearing echo as transport noise. An unsend produces exactly that shape with the opposite meaning: Apple withdraws only the sent copy, the echo keeps its text, and the empty row *is* the message. The retracted outbound row now wins and the echo is dropped, so the tombstone shows there too.
- **`bridge/mac/test_unsent.py`** (new) + one CI step, same shape as `test_group_photo.py`; `thread.test.ts` gains the self-thread case.
- **`scripts/demo/fake-imsg`** — passes `edited` / `retracted` through from a fixture row instead of hardcoding false, so the harness can render the tombstone. The shipped demo data is unchanged; the screenshot above used one extra fixture row locally.
- CHANGELOG (Unreleased); CLAUDE.md gets the invariant so the two columns are not conflated again.

## How it was verified

- [x] `bun test` is green (CI will check) — 324 pass; `test_unsent.py` 7 pass; `test_group_photo.py` pass; every Mac tool compiles
- [x] UI change → no layout change; the existing tombstone renders where an empty bubble did — screenshot above (harness, invented people)
- [x] Touches an invariant in `CLAUDE.md` → **new: an unsend is not an edit, and on macOS 26 it never sets `date_retracted`**

Against the live database, old bridge vs new, `--json --rich thread --chat` on the thread that showed the problem: comparing every field of every row, the only differences are the flags on the three rows in question, `edited` → `retracted`. Over every row with an edit date, a retraction date or a summary blob, the new decision yields 14 retracted and 8 edited, matching the column-level analysis exactly. After install, the thread shows 3 tombstones and 0 empty non-retracted bubbles. A real Undo Send performed on a self-thread during review reproduced the echo problem (the withdrawn text kept showing) and is now rendered as the tombstone.

The 7 tests use `plistlib`-built blobs in the shapes Apple writes (`rp` / `ec` / `otr` / `amc` / `ust`): `rp` means unsent; a real edit does not; missing, empty, non-plist, empty-`rp` and non-dict blobs do not; the macOS 26 unsend is retracted and not edited; the older unsend via `date_retracted` still works; a real edit is edited; an untouched message has no flags.

Note: this branch and #31 both add a line to `ci.yml` and to the top of the CHANGELOG; whichever lands second gets a one-line rebase from me.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,6 +17,8 @@ jobs:
         run: python3 -m py_compile bridge/mac/imsg bridge/mac/imsg-send bridge/mac/contacts bridge/mac/tcc-check bridge/mac/blip-check bridge/mac/blip-dispatch
       - name: group-photo lookup (no macOS needed)
         run: python3 bridge/mac/test_group_photo.py
+      - name: unsent messages are tombstones, not empty bubbles (no macOS needed)
+        run: python3 bridge/mac/test_unsent.py
       - name: shell scripts (bash -n + shellcheck)
         run: |
           bash -n bridge/linux/blip-shim scripts/blip-setup scripts/sync-bridge.sh bridge/mac/install.sh
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,17 @@
 
 ## Unreleased
 
+- **Unsent messages are tombstones again, not empty bubbles.** On macOS 26 an
+  Undo Send no longer sets `date_retracted`; Apple records it as an edit —
+  `date_edited` stamped, body cleared, the withdrawn parts listed as `rp` in
+  `message_summary_info`. The bridge read only the date columns, so every
+  unsend since January arrived as an empty bubble tagged "Edited". It now
+  reads the summary blob too: `rp` means unsent (and not edited), `ec` means
+  a real edit as before, and the old `date_retracted` path still works on
+  older Macs. In the conversation with yourself, where every message lands
+  twice, the withdrawn copy now wins over its echo instead of being folded
+  away as transport noise. Mac side: re-run the install one-liner so
+  `~/.blip/bin/imsg` picks it up.
 - **Group photos show up as photos again.** Messages stores a group's picture
   on an announcement row (`item_type = 3`). The bridge hides those rows so
   they never become empty unread bubbles, and the photo lookup joined through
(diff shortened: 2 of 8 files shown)


nixfred:
Merged as 5888d0d — thank you, and this one taught me something. I had "an unsend sets date_retracted" filed as a fact; on macOS 26 it simply is not true any more, and the `rp` list in `message_summary_info` being the only trace is the sort of thing nobody finds without dumping ~33,000 rows and counting. The self-thread inversion is the detail I would have missed: same shape as the echo case, opposite truth, and your test pins it.

Verified here: `test_unsent.py` 7/7 on my Linux box (the plistlib-built blobs make it macOS-independent, which is why it is in CI now), 329 tests green on the merged tree, `imsg` compiles and runs on my Mac with the new column in `base_select`. My database has zero unsent rows in the recent window, so the honest statement is: nothing changed for me and the tests carry the proof. Merged by hand; no rebase needed.

PR #33, opened 2026-09-04
Drop rows with neither a chat nor a handle: leftovers of deleted conversations

Fileri:
## What

A message row with neither a chat nor a handle no longer becomes a thread. Such rows are leftovers of deleted conversations; Messages shows them nowhere. Rows that have a handle are untouched.

## Why

Deleting a conversation removes its chat row and the `chat_message_join` rows, but Messages in iCloud keeps the message rows in sync regardless, and some carry no handle either (sent messages, older group announcements). Blip grouped them by their empty identity into **one thread with no name and no number**, dated by its newest member, that could not be opened. On one database: 21 such rows spanning 2018 to 2026, surfacing as a single nameless thread near the top of the list.

The obvious "require a chat join" filter would be wrong, and the data shows why: some SMS senders — banks, services, short codes — get a chat row that their messages are never joined to, and Blip shows them only because `chatKey()` falls back to the handle. Nine senders on the same database exist solely that way. So the rule is the narrowest one: no chat **and** no handle means no identity, and no identity means no thread.

## How

- **`collector.ts`** — `hasIdentity(m)` (`chatKey(m) !== ""`), applied in `fetchMessages()` where every message enters, so threads, toasts and the unread ledger never see such a row.
- **`collector.test.ts`** — a normal row, a handle-only row and an identity-less row in; the first two come out.
- CLAUDE.md's `chat:null` invariant gains the rule; CHANGELOG under Unreleased.

## How it was verified

- [x] `bun test` is green (CI will check) — 324 pass
- [ ] UI change → none; one row disappears from the list
- [x] Touches an invariant in `CLAUDE.md` → **`chat:null` exists** — extended, not changed

Against the live bridge, running collector vs patched, `--deep`: threads 306 → 305, the nameless thread is the one that goes; short-code threads 3 → 3; alphanumeric-sender threads 160 → 160.

Note: like #31 and #32, this adds a line at the top of the CHANGELOG (and an invariant line in CLAUDE.md); whichever lands later gets a one-line rebase from me.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+- **No more nameless conversation at the top of the list.** Deleting a
+  conversation removes its chat row, but Messages in iCloud keeps the message
+  rows, and some arrive with no handle either. Grouped by their empty identity
+  they became one thread with no name and no number that could not be opened.
+  Rows with neither a chat nor a handle are now dropped where messages enter
+  the collector; rows with a handle are untouched, since some SMS senders only
+  ever exist that way.
 - **Group photos show up as photos again.** Messages stores a group's picture
   on an announcement row (`item_type = 3`). The bridge hides those rows so
   they never become empty unread bubbles, and the photo lookup joined through
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -134,7 +134,11 @@ what it is handed. Keep it that way.
   true, which would have silently hidden the GA Power SMS from 99123 that took
   a day to make arrive at all. A default that hides messages is not a
   preference.
-- **`chat:null` exists.** Use `chatKey()`; never `String(m.chat)`.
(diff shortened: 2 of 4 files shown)


nixfred:
Merged as f151fac — thank you. This is the kind of fix I like most: the rule is exactly as narrow as the data demands and no narrower. "Require a chat join" was the obvious filter and you showed with numbers why it would have been wrong — nine senders on your database exist only through the handle fallback, and on mine the SMS short codes (the power company, the bank) are precisely those rows. No chat AND no handle is the right line.

What I verified here: your test on my Linux box, the merged tree at 329 tests green, and the collector against my live Mac before and after — 299 threads both times, no nameless one on mine, so for me this is a no-op that could not have hidden anything. Merged by hand (the CHANGELOG conflict was two insertions at the same anchor, kept both), so no rebase needed from you.

PR #34, opened 2026-09-04
iMessage-app cards read as text instead of a replacement character

Fileri:
## What

iMessage-app cards — Ask to Buy requests, Fitness sharing, Find My and friends — read as the sentence Messages itself shows for them, instead of a replacement character or an empty bubble. Everywhere: the conversation, the list preview, the toast, search. Link cards are unchanged.

## Why

These cards store `U+FFFD` as their message text and the real content in an MSMessage archive in `payload_data`. The bridge returned the text column, so every consumer showed `�`. On one family database that was 254 messages, 213 of them children asking to buy something, and the thread they live in read as a column of question marks.

Apple stores a plain-text fallback for exactly this case: `ldtext`, the sentence Messages shows when the receiving device cannot render the app. It is present on every Activity and Find My card and on 187 of 213 Ask to Buy cards; the rest carry the app's name.

## How

- **`bridge/mac/imsg`** — `message_text()` tries `_app_card_text()` first: for a balloon that is not the URL provider and has a payload, unarchive it and return `ldtext`, else the card's `caption`, else the app name `an`, HTML-unescaped (Apple escapes some sentences, "Photo &amp; Video", and leaves others raw) and capped at 500 characters. Anything unreadable → the row is left as it was.
- The link-card decoder's reference following becomes `_unarchive` / `_deref`, shared; `_deref` now also resolves an archived `NSDictionary`, which an MSMessage payload is at its root. Link cards decode as before.
- **`bridge/mac/test_app_cards.py`** (new) + one CI step, same shape as `test_group_photo.py`.
- CHANGELOG (Unreleased), README (Conversation features), CLAUDE.md (invariant).

No client change: bubbles, previews, toasts and search already take the bridge's text.

## How it was verified

- [x] `bun test` is green (CI will check) — 323 pass; `test_app_cards.py` 8 pass; `test_group_photo.py` pass; every Mac tool compiles
- [ ] UI change → none; the same bubbles carry real text
- [x] Touches an invariant in `CLAUDE.md` → **new: an iMessage-app card is text, in the payload, not in the text column**

Against a live `chat.db`, old tool vs new, `recent 600`: 577 ordinary messages unchanged, 17 link cards unchanged, the 6 app cards in the window went from the placeholder to sentences of 77–84 characters; none is a placeholder or empty afterwards. `--rich recent 300`: the `link` field is byte-identical on all rows, so the shared helper changed nothing for link cards. The thread that started this went from 6 placeholders to 0.

The 8 tests build MSMessage-shaped archives with `plistlib`: `ldtext` wins; `caption` when there is no `ldtext`; the app name when the card has no words; HTML entities decoded and a raw `&` preserved; a link card keeps its URL; no payload, garbage, a non-archive plist and a row without the two columns all leave the text alone.

Note: like #31–#33, this adds a line at the top of the CHANGELOG (and one CI step); whichever lands later gets a one-line rebase from me.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,6 +17,8 @@ jobs:
         run: python3 -m py_compile bridge/mac/imsg bridge/mac/imsg-send bridge/mac/contacts bridge/mac/tcc-check bridge/mac/blip-check bridge/mac/blip-dispatch
       - name: group-photo lookup (no macOS needed)
         run: python3 bridge/mac/test_group_photo.py
+      - name: iMessage app cards read as text (no macOS needed)
+        run: python3 bridge/mac/test_app_cards.py
       - name: shell scripts (bash -n + shellcheck)
         run: |
           bash -n bridge/linux/blip-shim scripts/blip-setup scripts/sync-bridge.sh bridge/mac/install.sh
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 ## Unreleased
 
+- **iMessage-app cards read as text instead of a replacement character.**
+  Ask to Buy requests, Fitness sharing, Find My and other app cards store
+  U+FFFD as their message text and the real content in an MSMessage payload.
+  Blip showed the � — or an empty bubble — in the conversation, the list
+  preview and the toast; on one family database that was 254 messages. The
+  bridge now returns the payload's `ldtext`, the sentence Messages itself
+  shows when the app cannot render the card, falling back to the card's
+  caption and then the app's name. Link cards are unchanged. Mac side:
+  re-run the install one-liner so `~/.blip/bin/imsg` picks it up.
 - **Group photos show up as photos again.** Messages stores a group's picture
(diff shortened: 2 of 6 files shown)


nixfred:
Merged as 1180faa — thank you. 254 question-mark bubbles on a family database, 213 of them children asking to buy something, is a very concrete reason for a change, and reaching for `ldtext` — the sentence Messages itself falls back to — is exactly right; anything else would have been Blip inventing words for Apple's apps.

What I checked, since this touches `message_text()` for every row: `test_app_cards.py` 8/8 here, the merged tree green, and on my live Mac the `link` field of every rich row in a 300-message window is byte-identical before and after — so the shared `_unarchive`/`_deref` changed nothing for link cards, matching what you measured. My database has no app cards in that window, so for me it is a no-op that costs nothing. The `payload_data not in row.keys()` guard for snippet rows was a nice catch. Merged by hand (the CLAUDE.md conflict was your new invariant landing in front of a line #33 had just reworded — kept both), no rebase needed.

PR #35, opened 2026-09-05
Security codes: toast a 2FA code as it lands, click to copy, hotkey to type it

adamgamble:
## What

macOS reads a one-time code out of an SMS and offers it to Safari's login form. This is Blip's version:

- A text that carries a code — "Your verification code is 483920", Google's "G-482913", the origin-bound `@example.com #493857` line — toasts the code as it arrives. **Click the toast to copy it.**
- A new IPC verb, `typecode`, types the pending code into whatever has keyboard focus. Bound to a key (README shows `SUPER + SHIFT + V`), it is the macOS one-click. `copycode` is the clipboard twin. Both sit behind `automation=on`.

## How

**collector.ts** — `extractCode()` wants a trigger word (code, passcode, PIN, OTP, verify, confirm, sign in…) and picks the 4–8 digit token nearest it; money, percentages, phone numbers and URLs never qualify. `selectCodes()` gates like `selectIncomingLinks`: inbound, newer than the watermark, never the self-thread, **never a group**, once per message through a persisted `code:` ring. Output gains `codes`.

**BarWidget.qml** — holds the newest code in memory for five minutes and nowhere else: not state.json, not argv, not the notification daemon's on-disk history (the toast carries the `transient` hint; Omarchy otherwise persists every body to `~/.local/state/omarchy/notifications/history`). The toast's default action copies instead of opening a thread. `copycode` feeds wl-copy through an env var and stdin.

`typecode` sends the code as `send_key_state` down/up events over Quickshell's Hyprland binding — the path Omarchy's universal paste uses — **not wtype**. Hyprland merges a virtual keyboard's keys with the physical keyboard's modifiers, so digits typed while Super+Shift were still held from the hotkey arrived as Super+Shift+<digit> and fired workspace binds. `send_key_state` bypasses bind processing. Letters and "-" (origin-bound codes) map to their xkb names, uppercase with SHIFT. Noted in CLAUDE.md so nobody reaches for wtype again.

## Verified

- 25 new tests (detector shapes, exclusions, nearest-trigger tie-break, selector gates); `bun test` 337 pass.
- Live: a Truist code ("Your one-time code is …") was detected, held, copied via `copycode`, and typed via the hotkey into a browser field. The typing path was also proven with a throwaway Quickshell config aimed at a scratch terminal by window address: `7aB-` arrived intact.
- Plugin log clean after deploy.

## Known limits

- Heuristic by design: "confirmed for 09/04 at 1030" would toast 1030. The cost is a toast nobody clicks.
- Respects Do Not Disturb like every other toast; the hotkey works regardless, since the code is held whether or not the card showed.

### DIFF
--- a/BarWidget.qml
+++ b/BarWidget.qml
@@ -3,6 +3,7 @@ import Quickshell
 import Quickshell.Io
 import qs.Commons
 import qs.Ui
+import Quickshell.Hyprland
 
 // blip — iMessage in the bar.
 //
@@ -641,8 +642,15 @@ BarWidget {
   // five minutes (never state.json — it is message text), toasts it, and on
   // request copies it (click the toast, or `copycode`) or types it into
   // whatever has keyboard focus (`typecode`, bound to a hotkey). The code
-  // reaches wl-copy and wtype through an environment variable and stdin —
-  // never argv, where any process could read it from /proc.
+  // reaches wl-copy through an environment variable and stdin, and the
+  // focused window as key events over Hyprland's socket — never argv, where
+  // any process could read it from /proc.
+  //
+  // NOT wtype. A virtual keyboard's key presses are merged with the physical
+  // keyboard's modifier state, so the digits typed while the user's fingers
+  // are still on Super+Shift became Super+Shift+<digit> — workspace binds.
+  // `send_key_state` (what Omarchy's universal paste uses) hands the key
+  // straight to the focused window without consulting binds.
   property var pendingCode: null         // {code, name, domain, ts} or null
   Timer {
     id: codeExpiry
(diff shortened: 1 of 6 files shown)


nixfred:
Merged as 3a09047 — thank you. This is the feature I did not know Blip needed until I read the first paragraph, and the shape is right: the code lives in one place for five minutes and the two verbs that move it are user-initiated and behind `automation=on`.

What I checked before merging, since this one handles credentials and synthesises keystrokes:

- **The transient claim holds.** Omarchy's notification history entries on my machine carry a `body` key on disk, and `plugins/notifications/Service.qml` reads the freedesktop `transient` hint and skips storage for it — so without that hint every 2FA code would have been written to `~/.local/state/omarchy/notifications/history`. You read the daemon; it shows.
- **`send_key_state` exists on this Hyprland** and refuses anything but a table, so the dispatch string is the right shape. The wtype trap you describe (digits arriving as Super+Shift+digit) is exactly the kind of thing that only gets found by typing a real code, and now nobody else has to.
- **Both verbs refuse cleanly with nothing pending** — `copycode` and `typecode` both answered "no code pending" live, which is the case that matters most for a key that types into whatever has focus.
- `extractCode` live on the merged tree: a Truist-style sentence, an order number, and an origin-bound line all answered as your tests say. 337 tests green.

One honest footnote for the record rather than a change request: the toast *body* still travels to `notify-send` as an argument, the way every Blip toast does — so the code is briefly visible in `ps` for the seconds the toast is up. The copy and type paths avoid argv exactly as you say. If that ever matters, the fix is the daemon's D-Bus API instead of `notify-send`, and it would apply to all toasts, not just this one.

Small thing I added in a follow-up: a README note that `SUPER + SHIFT + V` is free on stock Omarchy but may not be on someone's own bindings — it happens to be taken on mine.

PR #37, opened 2026-09-05
Keyboard cursor: scroll to follow it, stop at the ends, arrows in the window

Fileri:
## What

Keyboard navigation of the thread list follows the cursor. Arrow keys and j/k used to walk the selection below the visible rows — in the thread list, the search hits and the new-message picker — while the list stayed put. The list now scrolls to keep the cursor row in view, the cursor stops at the ends instead of wrapping, Up from the first row hands off to the search field and Down in an empty field hands back, and leaving a field or a conversation returns to the row you were on. The app window (SUPER+M) gets Up/Down/Enter list navigation, which it did not have at all.

Four commits, each reviewable on its own; the second and third are behaviour choices you may want to weigh separately.

## Why

`moveCursor`, `moveSearchCursor` and `moveNewCursor` changed an index and nothing else. The rows recoloured themselves, but `threadFlick.contentY` was never touched, so on a list longer than the panel the selection disappeared below the fold after a dozen presses. Once the list follows the cursor, three things around it that were harmless before become visible: the modulo wrap (Down on the last of 300 threads landing on the first now reads as a jump), `/` and `n` pressed deep in the list focusing a field that was scrolled out of view, and `back()` parking the list at the top after Esc regardless of where the cursor was.

The list is a multi-section Column — pinned grid, section headers, three Repeaters in one Flickable — so `ListView.positionViewAtIndex` is not available. That is the shape Omarchy's own audio and Tailscale panels have, and the fix is the helper they use.

## How

**1. `4d91bb7` — Keep the keyboard cursor row scrolled into view; arrows in the window**

- `scrollCursorIntoView()` on the view root: `mapToItem` the cursor row into `threadFlick.contentItem`, clamp `contentY` so the row sits inside the viewport with a `Style.space(6)` margin. Same code as Omarchy's `panels/tailscale/Panel.qml` `scrollItemIntoView`. Called synchronously from the move functions: a cursor move does not change the model, so the row is already laid out.
- Each of the four row delegates (thread row, pinned tile, search hit, contact hit) gets `readonly property bool hasCursor` and `onHasCursorChanged: if (hasCursor) root.cursorRow = <id>`. The move functions never translate indexes between `threads`, `pinnedThreads`, `unpinnedThreads`, `searchResults` and `newResults`; the row that matches says so. `cursorRow` is a `property Item`, so a destroyed row reads back as null.
- `root.cursorChat` (the chat id at `threads[cursor]`) replaces `threadIndex(modelData)` in the two thread-row colour bindings. `threadIndex()` scanned every thread, per row, per keypress — O(n²) on a 300-thread list; a string compare per row is O(n). `threadIndex()` had no other callers and is removed.
- `BlipWindow.qml`: `navCatcher` forwards Up/Down/Enter through a new `view.catchNavKey(key)`, sibling of `catchEscape`/`catchNavText`. It returns false while `editorActive`, so the compose, search and new-message fields keep their own arrow handling — the window's "normal editor behaviour" rule from the header comment is intact. Letters (j/k, Space) are deliberately not added in the window.

**2. `15e56df` — Stop the keyboard cursor at the list ends instead of wrapping**

- All three move functions clamp to `[0, length-1]` instead of `% length`. Omarchy's `Dropdown` clamps; Menu and Clipboard wrap — for a launcher-sized list a loop is fine, for 300 threads it is a jump. The wrap arrived with the `a98f59e` refactor and is not mentioned in the CHANGELOG or docs, so I read it as incidental. Revert this one commit if you prefer the loop.

**3. `d007d6c` — Up from the first row focuses the search field**

- `moveCursor(-1)` on row 0 calls `startSearch()`. `startSearch()` and `startNew()` now set `threadFlick.contentY = 0` themselves — the field sits above the rows, so any way into it (Up, `/`, `n`, IPC `find`) must start with it visible. Before, `/` pressed deep in the list focused a field you could not see.
- Down in an empty search or new-message field calls `listFromTop()`: leaves the mode and puts the cursor on row 0. The list is showing its top, so that is where the eye is. Esc keeps the old cursor instead. That is the `SearchableDropdown` round trip (Up at the top → field, Down from the field → first row).
- `exitSearch()` and `exitNew()` end with `Qt.callLater(scrollCursorIntoView)`: the thread rows are rebuilt when the mode flips, so the scroll is deferred a turn — the reason Omarchy's Bluetooth panel defers `positionViewAtIndex`. A no-op while the rows are gone, so `startNew()` calling `exitSearch()` is unaffected.
- `back()` keeps `threadFlick.contentY = 0` (a mouse user with no cursor lands at the top as before) and then calls `scrollCursorIntoView()` in the same deferred block, so Enter → read → Esc returns to the row you opened.
- Up from the first search hit or contact hit only scrolls the field back into view; their field already holds focus while the arrows walk the hits.

**4. `7e1678a` — ui.test.ts: pin the keyboard-cursor invariants**

- Four source-level checks in the existing "QML safety invariants" style: every move function clamps and ends in `scrollCursorIntoView()` (a modulo fails the test); rows register as `cursorRow` and `threadIndex()` stays gone; `startSearch`/`startNew` scroll to the top while `exitSearch`, `exitNew` and `back()` restore the cursor row in that order; `catchNavKey` yields to a focused editor before touching the arrows. A shared `qmlFunction()` helper slices a root-level function body up to its own closing brace, so the checks do not depend on the order functions appear in the file.

CHANGELOG (Unreleased) has one entry covering all four.

## How it was verified

- [x] `bun test` is green (CI will check) — 352 tests, 351 pass, 0 fail; the four new tests fail when the wrap is reintroduced in a copy
- [x] Any send/receive behaviour was exercised against **my own number only** — this change does not send
- [x] UI change → exercised by hand on a 305-thread list, bar panel and app window: Down/Up through the fold both ways, ends, Up → field → type → clear → Down → row 0, `n` → Esc from a deep cursor, Enter → Esc from a deep cursor, search hits Up past the first
- [ ] Touches an invariant in `CLAUDE.md` → no; the wheel-scroll and Flickable invariants are untouched (`MouseArea.onWheel`, `interactive`, `StopAtBounds` unchanged)

`qmllint` on the two files reports the same warnings as `upstream/main`, nothing new. Bindings re-evaluated per keypress: n string compares plus one `mapToItem`; before, n·n string compares.

Not changed, deliberately: the cursor is still an index into `threads`, as upstream has it, so a push refresh that re-sorts the list moves the highlight to whatever row now sits at that index. An identity-based cursor is the better model and a separate change.

### DIFF
--- a/BlipView.qml
+++ b/BlipView.qml
@@ -256,6 +256,10 @@ FocusScope {
   // so the move functions never translate indexes between the four models.
   // A destroyed row reads back as null (guarded QObject property).
   property Item cursorRow: null
+  // The cursor is kept while the search field holds focus (Esc and Down
+  // return to it), but a row must not LOOK selected while typing happens
+  // elsewhere — Up from the top and a click in the field both got here.
+  readonly property bool cursorShown: !searchField.activeFocus
   property bool pinToBottom: false   // scroll to the newest bubble once layout settles
   property bool bubbleFocused: false // a bubble's TextEdit has focus (text selection in progress)
   property string threadRunningChat: "" // chat owned by the current threadProc
@@ -1753,7 +1757,7 @@ FocusScope {
                   Layout.preferredWidth: Math.max(1, (pinnedGrid.width - pinnedGrid.columnSpacing * 2) / 3)
                   implicitHeight: pinnedColumn.implicitHeight + Style.space(4)
                   radius: Style.cornerRadius
-                  color: pinnedHover.hovered || hasCursor
+                  color: pinnedHover.hovered || (hasCursor && root.cursorShown)
                     ? Qt.rgba(root.foreground.r, root.foreground.g, root.foreground.b, 0.08)
                     : "transparent"
 
@@ -1957,7 +1961,7 @@ FocusScope {
                 Layout.fillWidth: true
                 implicitHeight: rowRow.implicitHeight + Style.space(root.splitView ? 20 : 12)
                 radius: Style.cornerRadius
-                color: rowHover.hovered || hasCursor
+                color: rowHover.hovered || (hasCursor && root.cursorShown)
                   ? Qt.rgba(root.foreground.r, root.foreground.g, root.foreground.b, 0.08)
                   : "transparent"
 
--- a/BlipView.qml
+++ b/BlipView.qml
@@ -730,7 +730,7 @@ FocusScope {
   }
   function moveNewCursor(dy) {
     if (newResults.length === 0 || dy === 0) return
-    newCursor = (newCursor + dy + newResults.length) % newResults.length
+    newCursor = Math.max(0, Math.min(newResults.length - 1, newCursor + dy))
     scrollCursorIntoView()
   }
 
@@ -841,7 +841,7 @@ FocusScope {
   }
   function moveSearchCursor(dy) {
     if (searchResults.length === 0 || dy === 0) return
-    searchCursor = (searchCursor + dy + searchResults.length) % searchResults.length
+    searchCursor = Math.max(0, Math.min(searchResults.length - 1, searchCursor + dy))
     scrollCursorIntoView()
   }
   function acceptSearchField() {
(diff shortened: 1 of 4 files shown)


Fileri:
Heads-up: two follow-ups build on this branch and will come as separate PRs once it lands, so review can stay small:

1. **Sidebar preview in the window** — resting the cursor on a row for ~250 ms shows that thread in the right pane, like Messages' sidebar, *without* marking it read; Enter, a click or typing commits it. No config key; split view only.
2. **PgUp/PgDn/Home/End in the lists** — thread list, search hits and the new-message picker, panel and window, with the same four keys Omarchy's menu/clipboard/emoji lists bind.

Both are done and running locally; I'm holding them so this PR stays reviewable on its own. #39 is independent of all of this.

PR #39, opened 2026-09-05
Read the conversation from the keyboard: bubble cursor, actions, paging

Fileri:
## What

Reading and acting on a conversation from the keyboard. Inside a thread the compose field holds focus, so until now the arrows only moved its caret and older messages meant the mouse wheel. Now, with the compose field empty:

- `↑` / `↓` select a bubble — newest first, kept in view, drawn with the theme's hover fill — from an empty compose field, or from the first / last line of a draft (the edge rule Omarchy's single-line list fields get for free). `↓` past the newest, or `Esc`, drops the selection and returns to the bottom with the bottom-stick re-armed.
- `Enter` on the selected bubble opens its first attachment, else its link card, else the first URL in its text. `Ctrl+C` copies its text — or, for a bubble that is only a picture, the picture itself as an image. `Ctrl+R` quotes it into the compose field as `> …`.
- `PgUp` / `PgDn` (Fn+`↑`/`↓` on a Mac keyboard) select the topmost / bottommost visible bubble, and page a screen further when the selection is already there — with text in the compose field too, since they move no caret. `Shift+PgUp` / `Shift+PgDn` step one bubble at a time. `Home` / `End` select the oldest / newest bubble — from an empty field, or once the caret already sits at its line's start / end.

With text typed, the arrows and Home/End keep moving the caret; Enter keeps sending; a queued file's Enter stays a send. The status line under the compose box no longer collapses, so "copied" and "sending…" do not shove the conversation, and only failures are red.

Five commits, each on its own.

## Why

A keyboard user could open a thread with `Enter` and then had to reach for the mouse to read anything above the fold, and for every action on a message. Omarchy's own panels are driven from the keyboard end to end, and the list side of Blip already is; the conversation side was not.

The selection is deliberately a *target* rather than a scroll position: it gives copy, open and quote something to act on, and it costs one index into `bubbles` — no model change, no new state on the Mac.

## How

**1. `5bbbcef` — a bubble cursor, paging, Home/End**
- `bubbleCursor` (index into `bubbles`) and `bubbleCursorItem`, registered by the row itself (`hasCursor` → `bubbleCursorItem = bubbleRow`), the pattern the thread list uses. `onBubblesChanged` clears it: a reload renumbers the rows.
- The highlight is one `Rectangle` beside `content` in the Flickable, positioned from the row's `y`/`height`. Bubbles are several parts (text, link card, attachment images), so a band behind the whole row beats a ring on each part, and no delegate carries a background. Its fill is `Style.hoverFillFor(foreground, accent)` — the theme's `hover-cursor-color` / `hover-cursor-fill-alpha`, which Omarchy's own rows use — rather than a hard-coded foreground at 0.08 (the thread list's rows still hard-code that; a follow-up).
- `scrollConversation(dy)` is now the one writer of `flick.contentY` — the wheel handler goes through it too — and therefore the one owner of the bottom-stick that gates the deferred push reload (invariant added to CLAUDE.md).
- `Esc` drops the selection first and leaves the thread second: the same two-step Esc a text selection in a bubble already has.

**2. `221f9f7` — Enter, Ctrl+C, Ctrl+R on the selected bubble**
- Only with an empty field and no queued file (`draftPath === ""`): a queued file's Enter is a send. Enter is otherwise free here — `send()` already returns on empty text.
- `openBubble()` reaches the same `openAttachment()` / `openLink()` a click does. `Ctrl+C` matches `StandardKey.Copy`, so Omarchy's universal `SUPER+C` (which the compositor forwards as Ctrl+C) works too, exactly like the existing `StandardKey.Paste`.
- The quote is plain text on purpose: iMessage's inline reply is not reachable through the bridge — no message GUID leaves the Mac in a bubble, and Messages' AppleScript dictionary has no reply-to.

**3. `c6ee163` — Ctrl+C on a picture bubble copies the picture**
- `copyBubble()`: text wins; a picture-only bubble copies its first image attachment. That rides the existing fetch queue (the Mac round trip a click makes): `enqueueFetch`'s boolean "open when done" becomes an action string (`""` / `"open"` / `"copy"`), and a copy job ends in `wl-copy --type <mime> < <file>` with path and type as arguments, never interpolated into the script.
- The copy fetches the *original* while the bubble draws the 1600 px preview the auto-fetch made; the inline source and metrics are left alone for a copy job, otherwise the picture re-decoded and re-laid-out under the reader's eyes (seen). HEIC/HEIF arrive as JPEG (`fetch.ts` `wantsJpeg`), so the clipboard type says `image/jpeg` for those.

**4. `7e2b625` — the status line keeps its height; only failures are red**
- The note renders a single space when empty, so nothing moves when "copied" appears. Progress and confirmations are dim; failures red — before, everything but "sending…" was red.

**5. `3e51720` — walk the bubbles with a draft in the field: edge rule, PgUp/PgDn, Shift**
- Edge rule: Up leaves the draft from its first line, Down from its last while a bubble is selected; the caret rectangle decides, since the compose box is multi-line. PgUp selects the topmost bubble with any part in view and, when the selection is already there, pages up a screen first; PgDn mirrors it with the bottommost, and past the newest leaves like Down. Only a wholly visible row counts as the edge (a sliver of the row above otherwise turned the next PgUp into a single step), and the page lands the selected row at the opposite edge — one row of overlap. `Shift+PgUp`/`Shift+PgDn` step one bubble at a time, draft or not. They move no caret, so this is the one way to walk the bubbles while a reply is half-typed (Fn+Up/Down on a Mac keyboard). `edgeVisibleBubble()` scans the Repeater's items, at most 80.

README rows, CHANGELOG (Unreleased), four invariant tests in `ui.test.ts` (`qmlFunction()` helper — same one #37 adds; whichever lands second gets a one-line conflict there).

## How it was verified

- [x] `bun test` is green (CI will check) — 350 pass
- [x] Any send/receive behaviour was exercised against **my own number only** — this change does not send
- [x] UI change → exercised by hand on Omarchy, panel and window: arrows up through a long thread and back down past the newest, Esc, Enter on a picture (opened), Ctrl+C on text then paste, Super+C on a picture-only bubble then paste into an image editor, Ctrl+R quote, PgUp/PgDn/Home/End, and the status line holding still through "copied"
- [x] Touches an invariant in `CLAUDE.md` → extends the wheel-scrolling invariant: every writer of the conversation's `contentY` goes through `scrollConversation()`

## Addendum (sixth commit): Enter on a link opens the share sheet, and the sheet takes the keyboard

Enter on a selected bubble whose text is a link went straight to the browser. A single keypress that opens an arbitrary URL, with no chance to see where it goes, is the wrong default for a messaging client. The share sheet already exists for exactly this: it names the host, shows the full URL, and offers Open / Copy / Send as a deliberate choice. Enter now goes there; a picture still opens directly (it is local).

The sheet was mouse-only, so it takes the keyboard: `1`/`2`/`3` pick, `↑`/`↓` move the highlight, Enter takes it, Esc closes. Hovering moves the same highlight, so mouse and keys agree. Any other key falls through to the compose field, so a sheet over a draft never blocks typing. **A sheet that opened by itself behaves the same** (a link you sent, a link that arrived while Blip was open, IPC): the sheet is the warning, with the host, the full URL and a button that says what Enter does. The one difference is a 700 ms grace after it appears, when Enter and digits still belong to the draft, so a link landing in the same instant as the Enter meant to send is never opened by it.

**Several links in one message.** Enter used to reach only the first. The sheet now takes every link of the message, in order, found with the same pattern and punctuation rule `linkify()` uses for the anchors the mouse clicks (plus Apple's card URL when the text lacks it). The same list arrives after a send and from the collector (`IncomingLink.urls` beside `url`; readers of `url` are untouched), through `BarWidget` and `shareLink()`, which takes one URL or a list. `←`/`→` and `‹ ›` step through them, the header counts "1 of 2", host / URL / QR follow. Stepping opens nothing, so it works on an auto-opened sheet too, but only with more than one link, so a draft keeps its caret keys.

**Found on the way, upstream's:** the sheet's `TapHandler`s took only a passive grab, so a click on the scrim, the card or a button also reached whatever lay under it (a thread row, a text selection, a link in a bubble). `gesturePolicy: TapHandler.ReleaseWithinBounds` takes an exclusive grab on press, so the tap now ends in the sheet. The compose field's Esc closes the sheet before it drops a bubble selection or leaves the thread (the audit's `shareUrl` check, on this branch too, so the rebase merges cleanly).

- [x] `bun test` — 353 pass; `collector.test.ts` covers `urls`
- [x] Verified live in the panel and the window: Enter on a link → sheet with the host; `2` copies; Esc leaves the draft untouched; a two-link message, sent and received, shows "1 of 2" and `←`/`→` / `‹ ›` step

### DIFF
--- a/BarWidget.qml
+++ b/BarWidget.qml
@@ -101,10 +101,12 @@ BarWidget {
   /** Newest arriving link → the share sheet on whichever surface is open. */
   function shareArrivingLink(link) {
     if (!link || !link.url) return
+    // Every link of the message: the sheet steps through them.
+    var urls = Array.isArray(link.urls) && link.urls.length > 0 ? link.urls.map(String) : [String(link.url)]
     var w = windowLoader.item
-    if (w && root.windowVisible && typeof w.shareLink === "function") { w.shareLink(String(link.url)); return }
+    if (w && root.windowVisible && typeof w.shareLink === "function") { w.shareLink(urls); return }
     var p = panelLoader.item
-    if (p && p.opened === true && typeof p.shareLink === "function") p.shareLink(String(link.url))
+    if (p && p.opened === true && typeof p.shareLink === "function") p.shareLink(urls)
   }
 
   function injectPanel() {
--- a/BlipView.qml
+++ b/BlipView.qml
@@ -2916,12 +2916,16 @@ FocusScope {
           }
         }
 
+        // Always one line tall, empty or not: a note that appears and vanishes
+        // must not shove the compose box and the bubbles around. Only failures
+        // are red; progress and confirmations are dim.
         Text {
           Layout.fillWidth: true
-          visible: root.note !== ""
-          text: root.note
+          text: root.note === "" ? " " : root.note
           textFormat: Text.PlainText
-          color: root.note === "sending…" ? root.dim : root.urgent
+          readonly property bool calm: root.note === "copied" || root.note === "sending…"
+            || root.note === "sent to LocalSend" || root.note.indexOf("attached") === 0
+          color: calm ? root.dim : root.urgent
           font.family: root.fontFamily
           font.pixelSize: root.fontCaption
           wrapMode: Text.WordWrap
--- a/BlipView.qml
+++ b/BlipView.qml
@@ -417,17 +417,6 @@ FocusScope {
     flick.contentY = Math.max(0, Math.min(max, flick.contentY + dy))
     flick.stick = flick.contentY >= max - 4
   }
-  /** Pixels a key moves the conversation from the compose field, 0 = not a
-   *  scroll key. PageUp/PageDown always page; Home/End only while the field
-   *  is empty — with text typed they keep moving the caret. */
-  function conversationStep(key, fieldEmpty) {
-    if (key === Qt.Key_PageUp) return -flick.height * 0.9
-    if (key === Qt.Key_PageDown) return flick.height * 0.9
-    if (!fieldEmpty) return 0
-    if (key === Qt.Key_Home) return -flick.contentHeight
-    if (key === Qt.Key_End) return flick.contentHeight
-    return 0
-  }
   /** Up/Down in an empty compose field walk the bubbles, newest first, and
    *  keep the selected one in view. Down past the newest drops the selection
    *  and re-sticks to the bottom, so the conversation follows new messages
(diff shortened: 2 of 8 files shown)


Fileri:
A sixth commit, `aac54f7`, from using the branch for a day: **Enter on a selected link now opens the share sheet, not the browser.** A keypress that opens an arbitrary URL with no chance to see where it goes felt like the wrong default for a messaging client, and the sheet already names the host and shows the full URL, so opening becomes a deliberate second step. A picture still opens directly.

That made the sheet need a keyboard: `1`/`2`/`3` pick, `↑`/`↓` move, Enter takes the highlight, Esc closes; hovering moves the same highlight. Other keys fall through to the compose field, so a sheet over a draft never blocks typing. A sheet that opened by itself (sent / arrived / IPC) behaves the same, with a 700 ms grace after it appears during which Enter and digits still belong to the draft, so a link landing as Enter is pressed to send is never opened by it.

Messages with several links offer them all, from Enter, from a send and from the collector (`IncomingLink.urls` beside `url`; readers of `url` untouched); `←`/`→` and `‹ ›` step, the header counts "1 of 2".

Two things found on the way, both yours to judge: the sheet's `TapHandler`s took only a passive grab, so a click on the scrim or a button also reached the row, bubble or link underneath — `gesturePolicy: ReleaseWithinBounds` ends the tap in the sheet. And this branch is based on `390df88`, so it carries the audit's `shareUrl` check in the compose field's Esc itself; the rebase onto main should merge cleanly.

353 tests, `collector.test.ts` covers `urls`. The PR description above has the details under "Addendum (sixth commit)".

nixfred:

Fileri:
For the record: this one was already merged as 68b87b3 before the rebase note went out, with the integration resolved in the merge commit. Thank you for taking that on rather than bouncing it back. The next branch, the sidebar preview, comes rebased on today's main.

PR #40, opened 2026-09-06
chats: follow Messages for re-keyed group pins and merged DMs

jefehoser:
## What

Favorites and the thread list follow Messages for two conversation shapes that used to split:

1. **A re-keyed group's pin stays visible.** Messages can give a group a new chat row (re-invite, iCloud shuffle) and keep the pin on the *retired* row's `group_id`. `imsg chats` only matched the live row, so that pin slot had no conversation. Pins now match every identifier in the cluster.
2. **A merged 1:1 is one thread.** Messages folds a phone SMS row and an email iMessage row under one `group_id` while keeping two `chat_identifier`s. Blip listed them twice, with the pin stuck on the stale SMS handle. Those rows now cluster like re-keyed groups, and opening the thread keeps both sides' history.

Mac side: copy `imsg` to `~/.blip/bin` (or re-run the install one-liner).

## Why

Messages.app already treats both as one conversation. Blip keyed pins and 1:1s on a single live id, so a pinned group vanished from Favorites and the open 1:1 showed yesterday's SMS while today's iMessages sat in a second, unpinned row.

This is the same cluster idea Astra #8 used for group *history*, applied to pins and to 1:1s that share `group_id`.

## How it was verified

- [x] `bun test` is green (CI will check) — 358 pass, including fold + `selectThread` alias tests
- [x] Mac-side unit tests (`python3 bridge/mac/test_cluster.py`, `python3 bridge/mac/test_group_photo.py`) — no live `chat.db`
- [x] Live `imsg chats` after deploying the Mac tool: every pin slot resolved, a merged 1:1 listed once under the live iMessage handle (metadata only; no conversation contents)
- [ ] Any send/receive behavior was exercised against **my own number only** — never a contact or a group — no send path in this change
- [ ] UI change → screenshot attached — Favorites tiles changed membership, not layout; no screenshot of a real conversation

Invariants: pins and merged 1:1s use the whole conversation cluster; `thread.ts` already keeps group alias rows (Astra #8) and now keeps merged-DM alias rows the same way; `MESSAGE_SRC` still hides announcements.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,8 @@ jobs:
         run: python3 bridge/mac/test_app_cards.py
       - name: unsent messages are tombstones, not empty bubbles (no macOS needed)
         run: python3 bridge/mac/test_unsent.py
+      - name: conversation cluster pins (no macOS needed)
+        run: python3 bridge/mac/test_cluster.py
       - name: shell scripts (bash -n + shellcheck)
         run: |
           bash -n bridge/linux/blip-shim scripts/blip-setup scripts/sync-bridge.sh scripts/demo/blip-shots bridge/mac/install.sh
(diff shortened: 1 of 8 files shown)


jefehoser:
Still loving Blip!

PR #45, opened 2026-09-07
The window's sidebar previews the thread under the cursor without marking it read

Fileri:
## What

In the app window, resting the sidebar cursor on a conversation shows it in the right pane, the way Messages' sidebar does. No Enter needed, focus stays in the list, so the arrows keep walking. A previewed thread is **not** marked read; only Enter, a click or typing, which put focus in the compose field, commit it. `→` steps from the list into the compose field and `←` from the start of the text steps back. Leaving the list for the search or new-message field ends a preview, so the pane is empty again, and the sidebar keeps one spacing whatever the pane shows. The bar panel is unchanged.

## Why

The window has two panes, and the sidebar cursor from #37 walked the left one while the right one stayed on whatever was opened last. Messages shows the highlighted conversation as you move, and that is what the split view is for. But a glance is not a read: Messages keeps the blue dot until you actually land in the conversation, and the phone's read state must not flip because a cursor passed by. So the preview is gated everywhere a read can be recorded.

Two things fell out of using it. `startSearch()` and `startNew()` left a preview up, or let its timer fire after focus had moved, so the pane showed the top thread beside an empty search field. And the sidebar column's spacing followed `inThread`, a rule meant for the popout where `inThread` hides the list; in the window it flipped with every preview, moving the search field 8 px (measured: y 18 with a thread shown, 26 without).

## How

- **`BlipView.qml`** — `peekTimer` (250 ms, restarted on every cursor move) calls `peekCursor()`, which shows the cursor row's thread with `peeking = true`. `markRead(chat, seen)` is the one caller of the host's `markThreadRead` and checks `peeking`; `commitPeek()` (compose field focus) clears the flag and marks read with the rendered snapshot's `seenTs`. `clearThread()` is the one place the pane is emptied; `back()`, `resetToList()` and `endPeek()` go through it. `endPeek()` runs from `startSearch()` and `startNew()`. `Keys.onLeftPressed` on the compose field steps back to the sidebar from the text's start, except while the share sheet is up, where `←` is the sheet's. The sidebar column keeps `Style.space(10)` in split view.
- **`BlipWindow.qml`** — exposes `peeking`; `→` from the list focuses the compose field.
- **`BarWidget.qml`** — `readingSurface()` ignores a window that is only peeking, so the collector is never told a peeked thread is being read.
- **`ui.test.ts`** — the read gate (every read path checks `peeking`), split-view-only + debounced + cleared on the way out, and the two follow-ups. **`CLAUDE.md`** gains the invariant "A peeked thread is not read"; CHANGELOG under Unreleased.

Rebased on today's main, after #39: `markRead` carries the `seen` timestamp #43 introduced, and the compose field's Esc keeps the three-way order from #39 (sheet, bubble selection, thread).

## How it was verified

- [x] `bun test` is green (CI will check) — 397 pass, on each of the four commits
- [x] UI change → verified live on Omarchy with an incoming DM: the dot stays while the thread is previewed, goes on Enter or `→`, and the phone follows (`push_read=thread`); Up from the first row to the search field leaves the pane empty; the search field holds still while previews come and go
- [x] Touches an invariant in `CLAUDE.md` → **new: "A peeked thread is not read"**

Note: like the other open PRs, this adds to the top of the CHANGELOG; whichever lands later gets a one-line rebase from me.

### DIFF
--- a/BarWidget.qml
+++ b/BarWidget.qml
@@ -391,8 +391,9 @@ BarWidget {
     var p = panelLoader.item
     if (p && p.opened === true && p.inThread === true) return p
     var w = windowLoader.item
-    // the window must be FOCUSED to count as being read (war room #25)
-    if (w && w.visible === true && w.focused === true && w.inThread === true) return w
+    // the window must be FOCUSED to count as being read (war room #25), and
+    // a thread merely peeked from the sidebar cursor is not being read either
+    if (w && w.visible === true && w.focused === true && w.inThread === true && w.peeking !== true) return w
     return null
   }
   function activeReadChat() {
--- a/BlipView.qml
+++ b/BlipView.qml
@@ -1804,6 +1804,9 @@ FocusScope {
     if (key === Qt.Key_Down) { moveCursor(1); return true }
     if (key === Qt.Key_Up) { moveCursor(-1); return true }
     if (key === Qt.Key_Return || key === Qt.Key_Enter) { activateCursor(); return true }
+    // Right = into the right pane: focus the compose field of the thread on
+    // screen (which commits a peek). Left in an EMPTY compose field comes back.
+    if (key === Qt.Key_Right && inThread) { composeField.forceActiveFocus(); return true }
     return false
   }
   function catchEscape() {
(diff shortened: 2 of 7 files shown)


nixfred:
Thanks — and thanks for turning #39 around so quickly.

This one lands as conflicting against current `main`. Could you rebase and push? Happy to review once it applies cleanly.

Fileri:
Rebased onto ba1d8f9 and pushed; applies cleanly now. The compose field moved into its Flickable underneath this, so the branch's one hook there (commitPeek on focus) and the Left handler were re-placed inside the new TextArea. Main is moving fast tonight — if it slips again before you get to it, say so and I will rebase once more.

PR #46, opened 2026-09-07
blip-setup pins the dedicated key to the enrolling machine over Tailscale

Fileri:
## What

`blip-setup` pins the dedicated Blip key to the machine it is enrolled from, when that machine reaches the Mac over Tailscale. The `authorized_keys` line reads `from="<this node's Tailscale addresses>",restrict,command="…/blip-dispatch"`, so the key is refused from anywhere else even with the private key in hand. Over a LAN the key is left unpinned rather than stranded on a DHCP address. A re-run of `blip-setup` replaces the key's line, so a changed address is one re-run away. No new setting: the pin follows the transport.

## Why

The key was already confined to `blip-dispatch` (no shell, no forwarding, no pty), but accepted from any address. The confinement protects the Mac, not the data: the bridge tools read every message and send as you.

**The threat this closes, precisely:** a copy of `~/.ssh/blip_ed25519` used from another machine — a home-directory backup, a sync, a dotfiles repo, a disk read outside the box. It does not address a compromised Linux box (the attacker is already at the pinned address) or a leaked everyday key (that is a shell on the Mac, and nothing in Blip can stand in for it). And the Blip key cannot undo the pin: it has no shell, none of the bridge tools touch `~/.ssh`, and `blip-setup` itself needs the everyday key to enrol anything.

Over Tailscale — where the README sends people — the address is a per-node identity that survives reboots and network changes, so pinning to it costs nothing and turns a leaked key file into a dead file. Elsewhere an address is not an identity, so nothing is pinned there; `docs/SECURITY.md` gets a short "hardening by hand" section with the one-line LAN pin and the Tailscale ACL that closes port 22 one layer down.

I first built this with a `key_from` setting (`auto | off | addresses`) and took it out again: it existed for LAN users on DHCP, who are exactly the users who never open `bridge.conf`, and half the code was validating input that no longer exists.

## How

- **`scripts/blip-setup`** — `blip_key_from()` takes the address the setup session arrived from (`SSH_CONNECTION` on the Mac, one round trip before enrolment): a `100.64.0.0/10` or `fd7a:115c:a1e0::/48` address is pinned, and widened to both of the node's addresses when it is among `tailscale ip`; anything else prints nothing. Step 4b replaces the key's line (`grep -vF` the key, append, `mv` under `umask 077`) and says in its ✓ line whether the key is pinned and to what.
- **`collector.test.ts`** — three tests run `blip_key_from()` for real with a `PATH` without `tailscale`: Tailscale IPv4 and IPv6 pin, `192.168.1.5` / `100.63.1.1` / `100.128.0.1` / `10.0.0.2` / `""` / `mac.local` do not, and the re-run replaces the line.
- **README / SECURITY.md / CHANGELOG** — the sentences that counted the bridge tools ("four", "five"; there are seven) say "the bridge tools" and describe the pin; SECURITY.md row 2 names it and a "Hardening by hand" section covers the LAN pin and the ACL.

## How it was verified

- [x] `bun test` is green (CI will check) — 393 pass
- [x] The exact remote enrolment command, run locally against a file holding another user's key plus a stale line for the Blip key, leaves the other key alone and ends with one pinned line for the Blip key. `bash -n` passes.
- [x] My own Mac has carried the Tailscale pin by hand since 2026-09-03; the bridge has run on it every day since.
- [ ] Touches an invariant in `CLAUDE.md` → no

Note: adds a line at the top of the CHANGELOG like the other open PR; whichever lands later gets a one-line rebase from me.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## Unreleased
 
+- **`blip-setup` pins the dedicated key to this machine over Tailscale.** The
+  key was confined to the bridge tools but accepted from any address, so a copy
+  of the key file still reached your messages. Over Tailscale the enrolled line
+  now carries `from=<this node's Tailscale addresses>` beside `restrict` and
+  `command=`, and a leaked key file is refused from anywhere else. Over a LAN,
+  where an address can change, the key stays unpinned (`docs/SECURITY.md` shows
+  the manual pin). A re-run replaces the key's line, so a changed address is one
+  re-run away.
 - **Reading a conversation can now clear it on your phone.** `push_read` in
   `bridge.conf` gained a documented middle setting and a status surface. The
   default, `all`, pushes to the Mac *only* on the mark-all gesture — so reading
--- a/README.md
+++ b/README.md
@@ -202,9 +202,10 @@ Linux side. If the Mac is asleep, the widget dims and says so.
 
 No server, no telemetry, no accounts. **Everything between the two machines
 travels inside ssh** — message text, attachment bytes, the push ping — on a
-dedicated key the Mac confines to Blip's five tools; nothing is ever sent in
-the clear. The full inventory of what touches
-disk on both machines is in [docs/PRIVACY.md](docs/PRIVACY.md) — short
+dedicated key the Mac confines to Blip's bridge tools — and, over Tailscale,
+to this machine's address; nothing is ever sent in the clear. The full
+inventory of what touches disk on both machines is in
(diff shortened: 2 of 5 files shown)


Fileri:
Converting to draft for a night: as written this pins to the one address the setup session came from, which is exactly right over Tailscale (stable per node) but a trap over LAN with DHCP, and a re-run of setup would not refresh it. Coming back with a bridge.conf key (key_from = auto | off | addresses), a re-run that replaces the key's line, and both Tailscale address families pinned when tailscale is present. Not worth your time until then.

PR #47, opened 2026-09-08
UTC timestamps, animated GIFs, multi-file sends, and a 2-4x faster bridge

zachwilke:
Seven commits: a correctness fix for timestamps, two media/compose fixes, three
performance changes, and one docs correction. Rebased on `main` (including the
compose Flickable and per-conversation drafts). CI-equivalent checks are green
locally: 420 `bun test`, all seven Mac-tool tests including `test_cluster`,
`py_compile`, `bash -n`. Deployed and running on my Omarchy box.

**Three things below need your call, not just review — please read those first.**

## What's in it

**1. Time crosses the bridge as UTC** (`688f2cb`)

Stamps arrived as the Mac's naive wall clock and were compared against the
Linux clock — the same string only while both machines sit in one timezone. Two
ways it lied. A Mac an hour ahead put every read mark ahead of every message,
so nothing ever counted as unread. And once a year, in the DST fall-back hour,
the Mac's own clock repeats: two messages an hour apart carry the SAME stamp,
so ordering, the watermark and "newer than the mark" stop meaning anything for
that hour. Every `a.ts < b.ts` in the collector assumes lexical order is
chronological order; only UTC actually provides it.

`imsg`'s JSON is now ISO-8601 UTC; the plain-text renders keep the Mac's clock
(`fmt_ts_local`), because a person reading `imsg recent` wants the time they
remember. Labels convert to the reader's zone at render, so day dividers still
break at your midnight.

Upgrading is covered from both ends, and both halves are load-bearing together:
`loadState()` migrates marks a pre-UTC release wrote, and stamps are normalised
at the two fetch doors so a Mac still on the old bridge keeps working.
Migrating the marks alone would sort every message below every mark (`" "` <
`"T"`) and silently empty the badge and the toasts.

The suite runs pinned to `TZ=UTC`, where all of this is invisible; the new
`timezone.test.ts` and `bridge/mac/test_wire_time.py` set their own zones and
are what actually cover it.

**2. GIFs move** (`1feab93`)

An animated GIF was flattened twice. `fetch.ts` treats any `image/*` under the
auto-fetch cap as a preview candidate and asks the Mac to resample it with
`sips`, which collapses an animation to one frame — a 1.4 MB GIF reached Linux
as a 198 KB JPEG. And a QML `Image` paints one frame whatever you hand it.
Animated mimes now skip the resample and render through `AnimatedImage`; stills
stay on `Image`, which is what applies `autoTransform` (the EXIF fall-back).
`AnimatedImage` honours `sourceSize` (measured), so the decode ceiling still
applies. `imageMetrics` also learned the GIF header — it reported 0x0 before.

**3. A message can carry several files** (`20f8bc1`)

Dropping five photos attached one and discarded four in silence: the handler
read `urls[0]` and the draft was a single path. Drafts are a list now, ship one
part per file, caption on the first only, service captured when the batch
starts so switching threads mid-send cannot push a later part onto a different
service. A failing part stops the batch and leaves the rest attached. Capped at
ten. Chips are one per row, for the implicit-width reason the received ones are.

Note the rebase: `main` now uses `drafts` for the per-conversation TEXT draft
map, so the attachment list is `attachDrafts` / `attachCount` /
`addAttachment(s)` / `removeAttachment` / `clearAttachments`. Two properties
named `drafts` would have been a duplicate-property error.

**4. Performance** (`0727e91`, `5b8b01e`, `934ee1f`)

Nothing was slow; everything paid startup. A bridge call cost ~133 ms while the
SQL underneath ran in ~1 ms: 16 ms ssh, ~25 ms `blip-dispatch` Python start,
~20 ms `imsg`'s own, 27 ms to open a 218 MB chat.db.

- The shim fired a *second* full ssh round trip before every call as a
  connectivity probe — 41 ms, ~31% — bought only for a friendlier error. ssh
  reports transport failure as 255 and every caller already treats 255 exactly
  like 69, so the probe is gone and the shim translates. This also retires the
  trap that came with it: the probe HAD to be `ssh -n` or it ate stdin and
  emptied `--file-stdin` payloads. Verified with a stub ssh that stdin still
  reaches the remote command, that a transport failure exits 69, and that other
  exit codes pass through.
- `imsg serve` answers many requests on one process with chat.db already open,
  re-dispatched through the SAME parser the CLI uses so there is no second
  command table to drift. Read-only queries only; attachment/avatar stream
  binary and would park the channel behind a 100 MB photo, `watch` blocks by
  design. `blip-bridged` holds two such channels behind a unix socket.
- A handle with no exact last-ten key was compared against all 442 saved
  numbers — 298 conversations meant 60,112 `_same_number` calls. Bucketed by
  last-seven-digits, which is sound rather than convenient: the rule only
  matches on a suffix with a seven-digit floor, so the last seven always agree.
  Verified identical to the old full scan over all 750 distinct handles in a
  real chat.db, zero mismatches.

Measured, warm medians: opening a conversation **223 -> 83 ms**, a poll
**160 -> 39 ms**, a deep poll **707 -> 342 ms**, a bridge call **132 -> 14 ms**.
The push debounce came down 250 -> 60 ms and the post-send reload 600 -> 250 ms.

Three "obvious" fixes the measurements talked me *out* of, all recorded in
CLAUDE.md so nobody retries them: replacing `cmd_chats`' 300-query N+1 with a
window function is **slower** (70 -> 107 ms; the per-chat lookups are indexed),
bundling/compiling the TypeScript changes nothing (44 vs 44 ms; a compiled
binary is worse at 53 ms), and `python3 -S` saves 2 ms.

**5. Docs correction** (`586a4de`)

`CLAUDE.md` said sending tapbacks needs SIP-off injection. That predates
macOS 26: enumerating Messages' menu bar over System Events finds real items —
`Edit > Tapback Message...`, `Reply to Message...`, `Edit Last Message...`.
That is the same shape `imsg-read` used to overturn "marking read is
impossible". **Nothing is claimed to work**; the note records what is still
unproven (the items read `enabled=false` from the background, which by our own
read-push finding is evidence of nothing; the item acts on the SELECTED
message, and selecting an arbitrary bubble from Linux is the open problem).
Recorded because a stale "not possible" stops people looking for years.

## Your call

1. **`blip-bridged` is a Linux-side daemon, and ROADMAP retired `blipd`.** I
   think this is a different animal — a transport multiplexer, not an
   application daemon; it holds no state, the leader BarWidget supervises it
   exactly as it already supervises the push watcher, and it is an accelerator
   that nothing depends on: no socket, no `socat`, a dead daemon or an
   unparseable frame, and `bridgeRun()` falls through to the ordinary one-shot
   path. But it is your architecture call, and I would rather you reject it
   than discover it later. The other two perf commits stand alone without it.

2. **`bridge/mac/imsg` is vendored, and I changed it a lot** (UTC, `serve`, the
   contact index). CONTRIBUTING says fixes to `imsg` are welcome here, but these
   arguably belong upstream in claude-on-mac, and `BRIDGE-VERSION` now names a
   pin the file no longer matches. `sync-bridge.sh` will flag the divergence
   and stage an `.upstream` copy rather than clobbering, which is the intended
   behaviour — but you may want these landed in claude-on-mac first and
   re-vendored instead.

3. **`socat` becomes a soft dependency** for the fast path only. Without it
   Blip behaves exactly as before, just slower; `blip-setup` says so rather
   than failing. Happy to drop it for a different client if you would rather
   not add one.

## Not verified

- The multi-file chip stack and the GIF bubble are covered by tests and the GIF
  is proven animating from instrumentation (`playing=true frameCount=18`, frame
  counter cycling, ten successive screenshots all distinct), but I have not
  photographed the chip stack.
- Tested against one Mac/Linux pair only (macOS 26.6, 218 MB chat.db, ~3,700
  contacts). The timing numbers are from that machine.
- Nothing here was tested by sending to another person — sends went to the
  self-thread or not at all, per CONTRIBUTING.

Happy to split this into separate PRs if the bundle is too big to review; the
perf commits are sequential and share `bridge/mac/imsg`, which is why they
arrived together.

https://claude.ai/code/session_015xxS9WCT7ifHhbDjYyGnHS

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
       - name: unit tests (collector, thread, tier2, ui invariants)
         run: bun test
       - name: Mac tools compile (python3 syntax, no macOS needed)
-        run: python3 -m py_compile bridge/mac/imsg bridge/mac/imsg-send bridge/mac/contacts bridge/mac/tcc-check bridge/mac/blip-check bridge/mac/blip-dispatch bridge/mac/imsg-read bridge/mac/calling_codes.py
+        run: python3 -m py_compile bridge/linux/blip-bridged bridge/mac/imsg bridge/mac/imsg-send bridge/mac/contacts bridge/mac/tcc-check bridge/mac/blip-check bridge/mac/blip-dispatch bridge/mac/imsg-read bridge/mac/calling_codes.py
       - name: group-photo lookup (no macOS needed)
         run: python3 bridge/mac/test_group_photo.py
       - name: phone matching, the way Contacts does it (no macOS needed)
@@ -27,6 +27,8 @@ jobs:
         run: python3 bridge/mac/test_cluster.py
       - name: stamps cross the bridge as UTC (no macOS needed)
         run: python3 bridge/mac/test_wire_time.py
+      - name: the serve channel exposes only read-only queries (no macOS needed)
+        run: python3 bridge/mac/test_serve.py
       - name: shell scripts (bash -n + shellcheck)
         run: |
           bash -n bridge/linux/blip-shim scripts/blip-setup scripts/sync-bridge.sh scripts/demo/blip-shots bridge/mac/install.sh
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,8 @@ jobs:
         run: python3 bridge/mac/test_unsent.py
       - name: conversation cluster pins (no macOS needed)
         run: python3 bridge/mac/test_cluster.py
+      - name: stamps cross the bridge as UTC (no macOS needed)
+        run: python3 bridge/mac/test_wire_time.py
       - name: shell scripts (bash -n + shellcheck)
         run: |
           bash -n bridge/linux/blip-shim scripts/blip-setup scripts/sync-bridge.sh scripts/demo/blip-shots bridge/mac/install.sh
--- a/BarWidget.qml
+++ b/BarWidget.qml
@@ -552,7 +552,14 @@ BarWidget {
   }
   Timer {
     id: pingDebounce
-    interval: 250
+    // 60 ms, not 250: this delay is paid on EVERY received message and on
+    // every send (Messages writing the row is itself a chat.db change, so the
+    // ping is what resolves the "Sending…" bubble). A burst still costs one
+    // fetch — messages in a burst land milliseconds apart, far inside 60 ms —
+    // but a single message no longer waits a quarter second before anything
+    // starts. Measured: the fetch it triggers is ~116 ms, so the debounce was
+    // more than twice the cost of the work it was coalescing.
+    interval: 60
     onTriggered: {
       var panel = panelLoader.item
       // Carrying the open thread's chat means a message landing in the
--- a/BarWidget.qml
+++ b/BarWidget.qml
@@ -509,6 +509,33 @@ BarWidget {
                               root.activeReadChat(), root.activeSeenTs())
   }
 
+  // ------------------------------------------------- the accelerator channel
+  // blip-bridged holds `imsg serve` channels open to the Mac, so a query costs
+  // the query instead of ~90 ms of ssh + two Python starts + opening a 218 MB
+  // chat.db. Measured end to end: opening a conversation 223 → 103 ms, a poll
+  // 160 → 38 ms. It rides here for the same reason the watcher does — the
+  // shell is the process that outlives every one-shot script, which is why
+  // Blip needs no daemon of its own — and only on the LEADER bar, or every
+  // monitor would hold its own pair of Mac processes.
+  //
+  // Nothing depends on it: collector.ts and thread.ts fall back to the plain
+  // one-shot ssh path whenever the socket is missing or a frame does not
+  // parse, so a failure here costs speed and nothing else.
+  Process {
+    id: bridgeProc
+    command: [root.home + "/bin/blip-bridged"]
+    running: root.leader
+    onExited: bridgeRestart.restart()
+  }
+  Timer {
+    id: bridgeRestart
+    // It exits on its own when idle; the poller keeps it busy in practice, so
+    // this is mostly the crash path. Slow enough not to spin if the binary is
+    // missing entirely (a setup that never installed it).
+    interval: 30000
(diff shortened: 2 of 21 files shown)


nixfred:
Thanks for this, @zachwilke. All seven commits are worth having, and the UTC-on-the-wire change fixes an open ROADMAP item (Mac-local wall-clock stamps compared against the Linux clock).

It no longer merges: main has moved a long way (2.5.0 is being cut now), and GitHub reports conflicts in BlipView, collector and thread. As one +2182/−461 branch it's also hard to review safely, because the parts carry very different risk. Could you split it into separate PRs, each rebased on current `main`?

1. **UTC timestamps** (688f2cb, with `timezone.test.ts` and `test_wire_time.py`). Highest value; please send this first.
2. **Animated GIFs** (1feab93)
3. **Several files per message** (20f8bc1)
4. **Bridge performance** (0727e91, 5b8b01e, 934ee1f). This one needs the closest security look: `blip-bridged` holds a channel to the Mac open, which changes what the confined ssh key does in practice, so please say in the description how it keeps the forced-command gate and the `ssh -n` rules in CLAUDE.md.
5. **Docs** (586a4de). #69 is already discussing outbound tapbacks, so this could go there.

Smaller PRs will also get reviewed and merged sooner.

zachwilke:
Fred — thanks for the split request. #47 is too large to review safely against current `main` (2.5.0 moved a lot; BlipView / collector / thread conflicted), so I rebased each slice onto `main` as its own PR:

1. UTC timestamps (highest value, first): https://github.com/nixfred/blip/pull/80
2. Animated GIFs: https://github.com/nixfred/blip/pull/81
3. Several files per message: https://github.com/nixfred/blip/pull/82
4. Bridge performance (persistent channel + confined SSH key — please give this one a close security look; also includes the Codex P1 `bridge.conf` allowlist before `_spawn()`): https://github.com/nixfred/blip/pull/83
5. Docs — tapbacks-out is half wrong on macOS 26, with a pointer to #69: https://github.com/nixfred/blip/pull/84

This mega-PR can be closed in favor of those five. I am leaving that to you / Zach rather than closing it from here.

zachwilke:
@nixfred Closing this mega-PR in favor of the five smaller ones above — thanks again for the clear split guidance. Looking forward to your review on those (UTC first is ready whenever you are).

zachwilke:
Superseded by #80 #81 #82 #83 #84.

PR #71, opened 2026-09-13
blip-setup: ssh -n, so the Enter prompts do not hit EOF

ianswope:
## What

Eight `ssh` calls in step 4 of `scripts/blip-setup` run a remote command without `-n`, so each inherits the script's stdin and forwards it to the Mac. When `blip-setup` is run with stdin from a pipe or a redirect, they drain it; the later `read -r -p` then hits EOF and, under `set -e`, the script dies at

```
  Press Enter to run the permission check:
```

with exit 1 — **after** the Mac install, the python3 probe and the dedicated-key enrolment have already succeeded. The setup is actually complete at that point, which is what makes it hard to read: it looks like the permission check failed, and in fact it never ran.

## Why

Hit it installing 2.4.0 on Omarchy. The install *had* worked — `blip-check` passed all three grants, `imsg recent 3` returned messages — but the script exited 1 with a message implying the opposite, so the natural next move is to re-run it and re-enrol the key for no reason.

The reachability probe on the same line already passes `-n`:

```sh
if ssh -n -o ConnectTimeout=6 -o BatchMode=yes -- "$host" true 2>/dev/null; then
  ssh -- "$host" 'mkdir -p ~/.blip/src'        # <- and seven more like it
```

so this reads as an oversight rather than a deliberate asymmetry.

None of the eight remote commands read stdin — `install.sh`, `blip-check`, and the `python3 -c 'import sqlite3, subprocess'` probe were each checked — so `-n` costs nothing. `ssh -O exit` is left alone: it runs no remote command.

## How it was verified

- [x] `bun test` is green — 479 pass / 0 fail locally (477 before, +2 from the new test)
- [x] `bash -n scripts/blip-setup` clean (shellcheck not installed here; CI will cover it)
- [x] **Regression test added.** `blip-setup: ssh never eats the script's stdin` asserts at the source level that every `ssh` running a remote command carries `-n`, and that the two prompts which would starve are still present. Reverting the fix makes it fail and name all eight offending lines.
- [x] **Mechanism reproduced in isolation**, with a stub `ssh` that drains stdin unless passed `-n`: the same two-line script exits 1 before the prompt without `-n`, and reaches it with `-n`.
- [ ] Not exercised end-to-end against a Mac — re-running `blip-setup` rewrites the `authorized_keys` line on a working install, and I did not want to do that to prove a stdin bug. The stub above isolates the behavior being fixed.
- [ ] No send/receive path touched — nothing was messaged, to my own number or otherwise
- [ ] No UI change, so no screenshot
- [ ] No invariant in `CLAUDE.md` touched


ianswope:
Worth adding: `CLAUDE.md` already carries this as a stated invariant, for the shims rather than the wizard —

> **The Linux shims' ssh preflight must use `ssh -n`.** A bare `ssh <mac> true` connectivity probe EATS STDIN, which silently empties `imsg-send --file-stdin` payloads. Fixed 2026-08-31.

Same mechanism, same failure shape ("the stdin you needed is gone, silently"), just landing on `read -r -p` in `scripts/blip-setup` instead of on a `--file-stdin` payload. So this reads less like a new rule and more like the existing one reaching the last place it had not been applied — which is also why the invariant test asserts it at the source level rather than just fixing the eight lines.

nixfred:
Merged in 7f89aea. Thanks @ianswope. The part that makes this bug nasty is that it hits after install and key enrolment have already succeeded, so a working setup looks broken and invites a pointless re-enrol. Your description of that was exact.

Checked independently before merging:
- A stub `ssh` that drains stdin unless it gets `-n` makes a `set -e` script exit 1 before its `read -p`. With `-n` it reaches the prompt.
- `install.sh` and `blip-check` read no stdin, so `-n` costs nothing there. `ssh -O exit` is correctly left alone.
- Your test run against main's `blip-setup` names exactly the eight lines.
- It merged cleanly on top of #72, #73 and #74. 481 bun tests and `bash -n` pass.

I added a CHANGELOG entry in the merge. You were right that this is the existing shim `ssh -n` rule reaching the last place it hadn't been applied.

PR #72, opened 2026-09-13
Every Text sink declares PlainText, so none can fall back to AutoText

ianswope:
## What

Eleven `Text` elements in `BlipView.qml` declared no `textFormat`. Qt's default is `Text.AutoText`, which sniffs the string and renders it as **rich text** when it looks like markup — and a rich-text sink will fetch a remote `` by itself.

Adds `textFormat: Text.PlainText` to all eleven, plus a UI invariant test that walks every `.qml` file and asserts each `Text`/`TextEdit` block declares its own format.

## Why

**This is not a live vulnerability, and I'm not claiming one.** Nothing reaches those sinks with a `<` today:

- `collector.firstUrl` / `collector.allUrls` both use `[^\s<>"']+`
- their QML mirrors (`BlipView.firstUrl`, `BlipView.allUrls`) do the same
- the one unfiltered entry is the IPC `share` path, whose only guard is `^https?://` — and that is gated behind `automation=on` and sits inside the same-user boundary `docs/SECURITY.md` already accepts as out of scope

The argument is about the sink, not the filter. **45 of the 46 sinks that already declare a format say `PlainText`** — the single `RichText` is the deliberate linkified bubble (`hasLink ? TextEdit.RichText : TextEdit.PlainText`). So these eleven were the unfinished edge of a rule the file otherwise follows everywhere, and the invariant "a Blip Text sink does not parse markup" was true by upstream filtering rather than by declaration.

Two of the eleven do render text that arrived in a message — `root.shareUrl` and `root.linkHost(root.shareUrl)` in the share sheet — which is what makes it worth closing by construction rather than by those regexes staying correct forever. Note `linkHost`'s own capture is `([^/?#]+)`, which does *not* exclude `<`.

## How it was verified

- [x] `bun test` is green — 478 pass / 0 fail locally (477 before, +1 from the new test)
- [x] **Invariant test added** in `QML safety invariants`: walks every `.qml`, finds each `Text`/`TextEdit` block, checks its *own* properties (nested children stripped) for `textFormat`. Reverting the eleven makes it fail and name exactly them — `BlipView.qml:2436, 2683, 2921, 2946, 3627, 3753, 3762, 3772, 3779, 3819, 3836`.
- [x] **No visual change is possible here.** `AutoText` and `PlainText` differ only when `mightBeRichText()` is true; of the eleven, seven render literals (`"loading…"`, `"↑"`, `"‹"/"›"`, `"SHARE LINK"`, the key hints) and the rest render filtered or locally-derived strings (`avatarInitials`, `modelData.day`, `modelData.label`, `linkHost`, `shareUrl`). The only input that would render differently is markup in a URL — which is precisely the case being fixed.
- [ ] **No screenshot.** QML needs `omarchy-restart-shell` to take effect and this is my daily-driver bar, so I did not restart the live shell to photograph a change that is render-identical by the reasoning above. Happy to attach one if you'd rather see it.
- [ ] No send/receive path touched — nothing was messaged
- [x] Touches an invariant in `CLAUDE.md` → see below


nixfred (approved):
Thanks @ianswope. This is the right kind of fix: a rule the file already followed in 45 of 46 places, now enforced by a test.

Checked on our side:
- Rebased cleanly on current main (e4aa3af). `bun test`: 478 pass, 0 fail.
- Put main's BlipView.qml back under the new test: it fails and names exactly the eleven lines from your description, 2436 through 3836. With your change it passes.
- Skipping the screenshot is fine. `Text.PlainText` and `AutoText` only render differently when `mightBeRichText()` is true, and for these eleven sinks that only happens with markup, which is the case being closed.

One non-blocking note: `own = body.replace(/\{[^{}]*\}/g, "")` strips one level of nesting. If a child block has its own nested block, a `textFormat:` inside it could count for the parent. That can only let a sink slip through, never flag a false one, so it doesn't need a change now.

Approved. Fred will merge.

nixfred:
Merged in 24d63bc and deployed; Blip came back healthy after the shell restart. Thanks again @ianswope.

Follow-up on the note in my review: dbb4573 changes the test to keep only characters at depth 1 during the brace walk, instead of stripping one level of nested blocks afterwards. `Text { Rectangle { textFormat: …; Item { } } }` used to pass even though the `Text` itself declared nothing; now it fails. It still names the same eleven lines against the pre-#72 file.

PR #73, opened 2026-09-13
A security code reaches wl-copy on stdin, not through the environment

ianswope:
## What

`copycode` passed the digits to `sh -c 'printf %s "$BLIP_CODE" | wl-copy'` through `Process.environment`. This execs `wl-copy` directly and writes the code to its stdin, closing stdin after the single write. Updates the matching `CLAUDE.md` invariant.

## Why

Using the environment instead of argv was already a deliberate step — but it trades one `/proc` exposure for a **longer-lived** one. `wl-copy` does not exit after a copy; it stays resident to serve the selection. So the variable stays readable in `/proc/<wl-copy>/environ` for as long as the clipboard holds the code.

Measured on Omarchy with a dummy value:

```
$ env BLIP_CODE=DUMMY123456 sh -c 'printf %s "$BLIP_CODE" | wl-copy'
$ pgrep -a wl-copy
783855 wl-copy                          # still alive, serving the selection
$ tr '\0' '\n' < /proc/783855/environ | grep BLIP_CODE
BLIP_CODE=DUMMY123456
```

That outlives both things the surrounding code is careful about: the toast that offered the code, and the five-minute window `copyCode()` enforces before it refuses an older toast's code. `noteCode()` already keeps the digits out of the notification body specifically so they are not persisted — this is the same concern one process along.

**Scope, honestly:** `/proc/<pid>/environ` is `0400` and owner-only, so this is the same-user boundary `docs/SECURITY.md` already accepts as out of scope — not a new hole, and I'm not calling it a vulnerability. What is wrong is the **lifetime**: a value the design says lives five minutes in memory was sitting in a long-running process's environment instead.

The replacement is the `stdinEnabled` / `write` / close pattern the codebase already uses for `qrProc`, `threadProc`, `previewProc` and the contact-details worker. It also drops the shell hop.

## How it was verified

- [x] `bun test` is green — 478 pass / 0 fail locally (477 before, +1 from the new test)
- [x] **Clipboard result is byte-identical**, no trailing newline either way:
  `env BLIP_CODE=X sh -c 'printf %s "$BLIP_CODE"'` and `printf %s 'X'` both give `44554d4d59393837363534`
- [x] **The leak is gone on the new path** — after `printf %s 'DUMMY987654' | wl-copy`, no `wl-copy` process carries the value in its environment; `wl-paste` returns it correctly
- [x] Test added: the copy `Process` declares no `environment:`, runs `wl-copy` directly, and closes stdin after one write
- [x] Test values were dummies, and the clipboard was cleared afterwards
- [ ] No send/receive path touched — nothing was messaged
- [ ] No visible UI change, so no screenshot
- [x] **Touches an invariant in `CLAUDE.md`** → named below, and updated in this PR


nixfred:
Thanks @ianswope, this is a good catch, and the scope note is exactly right. It's not a new hole, but the code was living far longer than the design says it should.

I checked it on Fred's machine (gus) before recommending a merge:
- Reproduced the leak with a dummy value. After `env BLIP_CODE=… sh -c 'printf %s "$BLIP_CODE" | wl-copy'`, the wl-copy process that stays running still has `BLIP_CODE` in its `/proc/<pid>/environ`.
- On the stdin path, no wl-copy process has the value in its environment.
- `bun test` on this branch: 478 pass, 0 fail.
- `onStarted: { write(...); stdinEnabled = false }` is the same pattern as `ContactDetails.qml`'s worker, and `copyValue` is still cleared `onExited`.

It's ready for Fred to merge.

PR #74, opened 2026-09-13
README: say how to remove Blip, and what it leaves behind

ianswope:
## What

Adds a **Removing Blip** section to the README, between the install/verify material and **Contact review**.

## Why

The README covers install and update but never removal, so the only way to learn what Blip put on disk was to read `blip-setup`. That matters more here than for a typical plugin: the install touches `~/bin`, two ssh files, three XDG directories and the Mac's `authorized_keys` — and one of the caches holds real media.

It also documents what *isn't* removed, and why:

- the `Host` block in `~/.ssh/config` is appended only when that host had none, and is harmless to keep
- the Mac's TCC grants belong to `/usr/libexec/sshd-keygen-wrapper`, not to Blip, so revoking them affects everything else you reach over ssh

Both are things a user should do deliberately rather than as a side effect of uninstalling, so they are called out as optional with the reasoning attached.

## How it was verified

Every path comes from `scripts/blip-setup` and the source, not from memory — and then was checked against a real install on both machines:

- [x] All documented Linux paths exist as described: `~/.config/blip`, `~/.local/state/blip`, `~/.cache/blip`, `~/.ssh/blip_ed25519{,.pub}`
- [x] **Four** shims, not three — `imsg`, `imsg-send`, `imsg-read`, `contacts`, all confirmed to be `blip-shim`. `imsg-read` is easy to miss.
- [x] `~/.cache/blip` subdirectories listed as they actually are: `att`, `avatars`, `linkpreview`
- [x] The `.pre-blip.<epoch>` backups `blip-setup` leaves behind are mentioned, so a displaced tool can be restored
- [x] On the Mac: exactly **one** `authorized_keys` line of eight matches `blip-dispatch`, so the documented `grep -v` removes that line and leaves the everyday key alone. (If the Blip key were the only line, `grep -v` exits 1 and the `&&` stops the `mv` — it fails safe, leaving the file intact.)
- [x] `bun test` still green — 477 pass / 0 fail
- [ ] Docs only: no code, no UI, no send path touched

One thing I noticed but did **not** change, since it is a different file and a different PR: `CLAUDE.md`'s shape diagram says the shims install as `~/bin/{imsg,imsg-send,contacts}`, omitting `imsg-read`. Happy to fix that separately if you want it.


README.md
```diff
@@ -513,6 +513,70 @@ rather than hang:
...
+
+Your everyday ssh key and its access are untouched — the line removed above is
+only the confined Blip key.
+
+**6. The Mac's privacy grants, optionally.** Removing Blip does not revoke
+them, and they belong to `/usr/libexec/sshd-keygen-wrapper` rather than to
+Blip — anything else you reach over ssh may depend on them, which is why they
+are not part of the steps above. To revoke anyway: System Settings ▸ Privacy &
+Security ▸ **Full Disk Access**, **Automation** and **Contacts**, removing the
+`sshd-keygen-wrapper` entry from each. See [docs/SECURITY.md](docs/SECURITY.md).
```
nixfred, line 578: One addition: opt-in mark-read (`blip-check --markread` / `imsg-read`) also grants **Accessibility** to `sshd-keygen-wrapper`, so a user who turned it on would otherwise leave that grant behind.

```suggestion
Security ▸ **Full Disk Access**, **Automation**, **Contacts** and, if you
enabled mark-read on the Mac, **Accessibility**, removing the
`sshd-keygen-wrapper` entry from each. See [docs/SECURITY.md](docs/SECURITY.md).
```

nixfred (commented):
Thanks @ianswope, this is exactly the right kind of docs PR. I checked every path against `scripts/blip-setup`, `bridge/mac/install.sh`, `collector.ts` and `linkpreview.ts`, and against a live install: the four shims (good catch on `imsg-read`), the `.pre-blip.<epoch>` backups, `att`/`avatars`/`linkpreview`, and the `blip-dispatch` match in `authorized_keys`, including the fail-safe when the Blip key is the only line. All correct. `bun test` is 477 pass / 0 fail. There's one small suggestion inline.

nixfred:
Merged in 2fb0094, thanks @ianswope. I folded in the Accessibility suggestion, added a CHANGELOG entry, and took you up on the offer about `CLAUDE.md`: the shape diagram now lists `imsg-read` among the shims, so no separate PR needed.

PR #75, opened 2026-09-13
A 3-4 digit short code is a DM you can reply to, not a group

ianswope:
## What

`isGroupChat()` called anything that is not a phone or an email a group, and its idea of a phone was five digits or more. A carrier short code shorter than that — `611`, or T-Mobile's `2536` — fell through to "group", so the conversation opened read-only with *group id unknown* and there was no way to reply `STOP` or `HELP`.

Closes the open ROADMAP item, which asked for `{3,15}` **"once tested against a real one."** Tested against `2536`, the only 3–4 digit code in my `chat.db`.

## Why it is four changes, not one

The rule turned out to live in four places:

| Site | Role |
|---|---|
| `collector.ts` `isGroupChat()` | is this a group? |
| `collector.ts` `sendServiceForMessages()` | a failed iMessage to a bare number is really SMS — true of a short code too, so it widens the same way |
| `BlipView.qml` `isGroupId()` | comment says *"Same rule as collector.isGroupChat()"* |
| `BlipView.qml` `isSendable()` | a DM target is a phone or an email — **this is the one that makes the thread read-only** |

Widening only the collector would have split them: a DM in the ledger, a group in the panel — still read-only, and now inconsistent about why. So all four move together, and a test pins every copy to one bound.

Writing that test taught me something worth flagging: it has to compare the **open** form. A regex matching only `{n,m}` skips `{5,}` entirely rather than flagging it, so reverting one site read as *"no rule here"* and the test passed. It now matches `{(\d+),(\d*)}` and fails correctly. (I checked because of your note on #72 — this was the same shape of mistake.)

## The new upper bound

`{5,}` had none. E.164 caps a real number at 15 digits, so a longer digit string is not a phone: it stays a group. That is the safe direction — the invariant is that an unknown shape must never become a DM target. Nothing in this `chat.db` exceeds 12 digits (distribution: 4×1, 5×10, 6×3, 11×85, 12×1).

One consequence worth naming: a short code is now **sendable**. Some are receive-only and a reply will fail — but that surfaces as the Not-Delivered bubble with a reason added in 2.5.0, which seems better than a thread that cannot be typed in at all. Say the word if you would rather it stay read-only.

## How it was verified

- [x] `bun test` — **488 pass / 0 fail** (477 before; +11 across the two new test groups)
- [x] Classification checked against real data: `2536`, `611`, `99123`, `63692` → DM; phone, email, `chat<digits>`, 32-hex and the `+1...(filtered)` shape → unchanged; 16 digits → group
- [x] Reverting `collector.ts` fails the behavior test naming `2536`
- [x] Reverting **one** QML site fails the drift test (`Expected: 1, Received: 2`)
- [x] ROADMAP item ticked, with what was done and why the bound is what it is
- [ ] Nothing was sent — no reply was attempted to `2536` or any other code
- [ ] No visible layout change, so no screenshot
- [ ] No `CLAUDE.md` invariant changed: *"`isGroupChat()` is 'not a phone/email' — never a positive regex on one shape"* still holds; only the definition of "a phone" was corrected


nixfred:
Merged, thanks @ianswope! a 3–4 digit short code is now a DM you can reply to. The test that makes every copy of the rule use the same {3,15} limit is what makes this safe to keep. It's in main as of 186b536 and deployed on my machine. I merged it together with nine other PRs, and all 526 tests pass on the combined tree.

PR #76, opened 2026-09-13
No toast for the conversation you already have open

ianswope:
## What

`selectToasts()` gains a fourth gate: the conversation being read right now. Closes the open ROADMAP item.

## Why

The function documents itself as *"Gated three ways"* — inbound and newer than the watermark, sender allowlisted, not already toasted. Its call site never passed the one piece of state that was already in scope:

```ts
const toast = selectToasts(msgs, state.watermark, loadAllowlist(), state.toasted);
```

Every poll while a thread is open carries `activeReadChat()`, and `collect()` uses it a few lines earlier to advance `readMarks` and clear `exactCounts`. So **the same run that suppressed the badge still fired the notification.** The two disagreed: a toast announcing there is something to read, with nothing left to click through to.

ROADMAP files this as the *bar popout case*. Reading the code, it is **both surfaces** — `readingSurface()` returns the popout *or* the focused app window, and neither reached the toast path. Only the security-code filter is applied at `fireToasts`.

## Two decisions worth reviewing

**Aliases are included.** `readChat` goes in together with `aliasesOf(chatAliases, readChat)`, the same fold the read marks directly above already do. A message arriving under a re-keyed group's retired chat row is the same conversation on screen, and `CLAUDE.md` already says a read of a canonical conversation covers its aliases.

**No second opinion about what "being read" means.** I deliberately did not re-check loading/failed/peeked state in the collector. `BarWidget.activeReadChat()` already refuses a thread that is still loading (Codex #4), one whose load failed (Astra A#2), and one merely peeked from the sidebar cursor — so an id that reaches here has cleared all of that. Duplicating those rules in TypeScript is exactly how the QML and TS copies drift, which is the bug I just fixed in #75. The parameter's contract is stated in the doc comment instead.

The parameter defaults to `[]`, so every existing caller and test is unchanged.

## How it was verified

- [x] `bun test` — **488 pass / 0 fail**
- [x] Four new tests: the open conversation is silent; **every other conversation still toasts while it is open** (the gate must not be a blanket mute); an alias of the open conversation is silent too; nothing open behaves exactly as before
- [x] Reverting `collector.ts` fails the first of those
- [x] ROADMAP item ticked, noting it was both surfaces rather than only the popout
- [ ] **Not exercised end-to-end.** That needs a message to arrive from an allowlisted sender while that exact thread is open, and `CONTRIBUTING.md` rightly says never to test against real people — so this is unit-tested at the seam rather than staged with a live contact. The wiring it depends on (`activeReadChat()` reaching `collect()`) is pre-existing and unchanged.
- [ ] No UI change, so no screenshot
- [ ] No `CLAUDE.md` invariant changed — the read-mark and alias rules it leans on are quoted above, not modified


nixfred:
Merged, thanks @ianswope! the conversation you're reading no longer toasts, and neither do its aliases. Spotting that the same poll was already clearing the badge is what made this a real bug and not just noise. It's in main as of 186b536 and deployed on my machine. I merged it together with nine other PRs, and all 526 tests pass on the combined tree.

PR #77, opened 2026-09-13
README: the allowlist and mutelist examples now parse

ianswope:
## What

The allowlist and mutelist examples were fenced ` ```jsonc ` with a path comment above the object:

```
// ~/.config/blip/allowlist.json — re-read every poll, no restart
{ "allow": ["+15551234567", "them@icloud.com"] }
```

`loadAllowlist()` and `loadMutelist()` both use strict `JSON.parse` inside a `try` that returns `[]` on failure. **So the block, copied as shown, configures nothing** — and says nothing about it.

Measured:

```
$ printf '%s\n%s\n' '// ~/.config/blip/allowlist.json — re-read every poll, no restart' \
                    '{ "allow": ["+15551234567", "them@icloud.com"] }' > allowlist.json
$ bun -e 'import {loadAllowlist} from "./collector.ts"; console.log(loadAllowlist("allowlist.json"))'
[]
```

## Why it matters more than a normal doc typo

The failure is invisible in exactly the direction that hides it. An empty allowlist and an unparseable one behave **identically**: every message still counts on the badge, still shows in the list, and no toast ever fires. Someone who set this up from the README and got no notifications has no error to search for, a file that looks correct, and a plugin that is working perfectly in every other respect.

I hit this setting up my own allowlist — I only avoided it because I checked the loader first.

## The change

- Comment moves into the prose; fences become ` ```json `, so both blocks are paste-ready
- The toast paragraph now states it is strict JSON, that a parse failure is silently an empty list, and to check with `jq . ~/.config/blip/allowlist.json` before anything else
- It also says to list **every handle a person messages from** — a contact with an iCloud address *and* a phone needs both lines, or they go quiet whenever they switch. Not obvious from a single-handle example, and it silently halves your notifications rather than breaking anything.

The `shell.json` example above these two keeps its ` ```jsonc ` fence — that file is Omarchy's, not Blip's, and I did not want to assert anything about how it is parsed.

## How it was verified

- [x] `bun test` — **485 pass / 0 fail** (484 before, +1)
- [x] Test added: writes each README ` ```json ` block to a file and loads it through the **real** `loadAllowlist` / `loadMutelist`, so an example that stops parsing fails CI instead of going quiet
- [x] Re-introducing the `//` line makes that test fail with `Expected: > 0, Received: 0`
- [ ] Docs plus one test; no runtime code touched


nixfred:
Merged, thanks @ianswope! the allowlist and mutelist examples now actually parse. A config that silently does nothing is the worst kind of bug to find, and the test reading them straight from the README keeps it fixed. It's in main as of 186b536 and deployed on my machine. I merged it together with nine other PRs, and all 526 tests pass on the combined tree.

Issue #78, opened 2026-09-13
Deleting a message: does it need SIP off, and do you want it?


ianswope:
## The ask

Two actions on a bubble: **Delete…** (local, recoverable) and **Undo Send** (retract for everyone, ≤2 min, own messages only). Blip already *reads* both states and renders them correctly; it just cannot cause either.

Before I write any of it — **is this in scope, and is the mechanism below the one you'd want?** This touches the confined key and adds a destructive action, so it seemed wrong to arrive with a PR.

**To be clear about what this is:** unlike #69, I have *not* built this. I have not probed System Events, deliberately — `CLAUDE.md` says never to fire a consent prompt the user did not ask for, and #36 is a live example of an unanswered prompt being recorded as a denial that macOS 26 then refuses to undo. So this is a design proposal grounded in your existing code, not a report of something working.

## Why I think it does not need SIP off

`README`, `ROADMAP` and `CLAUDE.md` all file outbound mutation under *"requires code injection, requires SIP off."* That is true of the BlueBubbles approach. It is not obviously true of the approach Blip **already uses**: System Events driving Messages under the Accessibility grant, SIP on, which is exactly how `bridge/mac/imsg-read` clicks **Conversation ▸ Mark All as Read**.

#69 makes the same argument for tapbacks and reports it working. Deletion looks like the same family.

**The open technical question, stated honestly:** `imsg-read` drives a *menu bar* item (`menu 1 of menu bar item "Conversation" of menu bar 1`), which is the easy case. Delete and Undo Send live on the balloon's context menu / AX action list, which is the harder one — you must first identify the right balloon in the transcript. #69 reports balloons expose their action row as AX custom actions, but I have not confirmed that Delete is among them, and I am not going to guess in an issue.

## The part I am most confident about: verification

`imsg-read`'s hard-won lesson generalises cleanly here —

> A disabled menu item is NOT evidence of anything; chat.db is the referee.

Both outcomes are observable in `chat.db`, so neither action has to trust the UI:

| Action | Signal |
|---|---|
| Delete… | a row appears in `chat_recoverable_message_join` |
| Undo Send | `date_edited` set, body emptied, `rp` in `message_summary_info` — what `_edit_flags` already reads |

So the tool can do what `imsg-read` does: count before, click, wait for the store to settle, count after, and report failure when the number did not move — whatever the menu claimed.

[...]

Issue #36, opened 2026-09-05
macOS 26.6.2: sshd-keygen-wrapper → Messages Automation cannot be enabled


jacobaross:
# macOS 26.6.2: sshd-keygen-wrapper → Messages Automation cannot be enabled

## Versions

- Blip manifest: 2.3.2; pinned source commit `a0031784de1632ba14f9a3723f2d03429b4e02d4` (post-release main, not the v2.3.2 tag).
- Gateway: Apple Silicon; macOS 26.6.2, build 25G83, verified with `sw_vers`.
- Same symptom before the OS update on macOS 26.5.2, build 25F84.
- System Integrity Protection enabled, including after the update.
- Linux Omarchy/Quickshell client over private SSH; plugin deliberately left disabled at the Mac permission gate. Exact desktop versions not captured in this report; the failure occurs in the Mac-side permission check before enabling the UI.

## What happened

The SSH bridge can authenticate and run the confined Blip tools, but the documented Automation grant cannot be enabled. In System Settings → Privacy & Security → Automation → sshd-keygen-wrapper → Messages, entering the Mac password leaves the switch off with no visible error.

Before the OS update, `blip-check` through the dedicated restricted key reported successful Messages database and Contacts access, but the Messages Automation check timed out. No test message was sent. The optional System Events/Accessibility capability is intentionally not enabled.

Quitting/reopening System Settings and retrying did not help. Updating macOS and logging back into the desktop did not resolve the reported switch behavior. A read-only check after the update found the same denied Messages permission record, unchanged from before the update:

```text
service: kTCCServiceAppleEvents
client: /usr/libexec/sshd-keygen-wrapper
client_type: 1
auth_value: 0
auth_reason: 9
indirect_object_identifier: com.apple.MobileSMS
```

The post-update evidence above is the user's UI observation plus the saved TCC record, not a new send test.

## Steps to reproduce the setup failure

1. Install the Mac bridge tools from the pinned commit and use the dedicated restricted SSH key.
2. With the Messages user's Mac desktop logged in, run `blip-check` through SSH.
3. Attempt the documented Messages Automation grant, including the System Settings switch after the initial check times out.
4. Authenticate when prompted; observe that the switch remains off.

Expected: a usable normal macOS consent flow allows the SSH bridge to control Messages.

Actual: reads pass, but the required Automation permission cannot be enabled, so installation remains paused.

## Diagnostic detail / question

[...]

PR #79, opened 2026-09-14
feat: save new contacts and delete individual messages

Kb2uka:
## Summary

Hi Fred — you may know me as **@StudioXRadio on X**. This adds two actions to Blip: save an unknown sender as a new Mac contact, and delete one stored message after confirmation.

Contact creation keeps the selected sender's number or email, checks for duplicates using native Contacts predicates plus the bridge's regional phone matcher, and verifies the saved card. Message deletion preserves the original row ID, GUID, and chat identity, selects the exact bubble through Accessibility, and verifies that it left live history. Both use stdin for requests; direct database connections remain read-only.

The setup wizard installs the new helpers and shims. [Mac action setup](https://github.com/Kb2uka/blip/blob/feat/contact-save-message-delete/docs/MAC-ACTIONS.md) documents the approval clicks, hidden-file picker, optional `blip-check --mutations`, first-use Recently Deleted notice, and recovery behavior. No messages are sent by the setup checks.

## Root cause

The follow-up audit reproduced a phone-format validation bug: a leading parenthesis was refused before normalization. All four validators now accept a leading parenthesis, including contact review's Mac resolver. A separate live locked-desktop failure exposed a cycle in the Messages accessibility graph; traversal now visits each element once while retaining its node/depth limits, allowing the existing open/unlock guidance to surface.

The later review identified a reproducible failure path for `unavailable`: a nil/throwing native Contacts predicate aborted inspection before AddressBook lookup and regional phone matching. A synthetic framework regression now reproduces that path. The native lookup failure now falls back to the existing bounded AddressBook exact checks, allowing regional candidates to reach the Python matcher. AddressBook errors still block creation, and exact email duplicates remain checked. The contributor's original helper hashes match `e17d1327`; the exact native error on their macOS 26.3 device was not independently captured.

## Proof

Synthetic UI captures from the real QML components:

| Conversation | Delete confirmation |
|---|---|
| Conversation | Delete confirmation |

| New contact | Review | Saved |
|---|---|---|
| New contact | Review | Saved |

- Identity regression: 0/2 before, 99/99 focused tests after.
- Lost native acknowledgement, denied Contacts access, and large Unicode transport regressions failed before their fixes and passed after.
- September 13, macOS 26.5.2: two authorized disposable messages were deleted through live Messages and verified absent from live history; the operator also confirmed their disappearance in Blip. This exercised the native deletion path; the screenshots above remain synthetic QML captures.
- One synthetic contact was saved and read back exactly; duplicate full phone, regional phone, and case-varied email were refused. The exact fixture card was removed afterward.

- Follow-up regressions: parenthesized input failed in the Linux form and Python transport before the fix; the Mac resolver regression also failed before its validator was aligned. All now pass.
- Swift cycle regression failed as `unsupported` before the fix; self-cycle, shared-child, exact 2,000-node bound, over-limit and depth-limit cases now pass.
- Native follow-up on macOS 26.5.2: created/read back a synthetic contact; exact, local and parenthesized duplicates all refused as `duplicate`; exact fixture cleanup verified.
- September 14, macOS 26.5.2: the operator confirmed another deletion through the installed PR after unlocking the Mac. One attempt returned open/unlock guidance before a successful retry. The contributor has tested contact saving only on macOS 26.3; they have not granted Accessibility or tested deletion.

- Native predicate failure regression: 14 passing / 2 failing before; 16 passing / 0 failing after. Covers nil/throwing predicates, regional candidate collection, exact phone/email duplicates, denied access, and failed fallback. A drift check compares the phone-format grammar across all four validators. This follow-up uses synthetic fixtures and performs no real Contacts or Messages mutations.

## Gates

Latest Contacts fallback follow-up (`423496c`): 505 TypeScript tests and 102 Python tests passed on a remote Linux host; Python compilation, Bash syntax, and ShellCheck passed. On macOS 26.5.2, 38 Python/Swift tests, Swift typechecking and JavaScript compilation passed. The later test-only addition also passed the full 505-test TypeScript suite; production code was unchanged. [Hosted CI awaits maintainer approval](https://github.com/nixfred/blip/actions/runs/34899865424).

- Follow-up: 500 TypeScript tests passed on a Linux test host.
- 102 Python tests passed on Linux; the macOS-only Swift harness skipped there and passed on the Mac.
- 42 Python tests, the Swift traversal regression, and Swift typecheck passed on macOS 26.5.2.
- Python compilation, Bash syntax, and ShellCheck passed.
- Original feature: wide/narrow QML save, confirmation and completion captures passed. The follow-up changes no QML layout.
- Native contact follow-up passed with exact fixture cleanup; no real message deletion was automated during this audit.
- Hosted CI requires maintainer approval for this fork; it is not claimed green.

## Review rounds

| Round | Seat A | Seat B | Accepted | Rejected | Needs a human |
|---|---|---|---|---|---|
| 1 | 5/5 (0) | 4/5 (9) | 5 | 4 | 0 |
| 2 | 5/5 (0) | 4/5 (1) | 1 | 0 | 0 |
| 3 | 5/5 (0) | 4/5 (2) | 2 | 0 | 0 |
| 4 | 5/5 (0) | 4/5 (1) | 1 | 0 | 0 |
| 5 | 5/5 (0) | 5/5 (0) | 0 | 0 | 0 |

Seat A used a substitute after the configured reviewer could not authenticate. No open findings remain. Later rounds corrected install inventories and permission guidance; action logic was unchanged after round 2.

Follow-up audit (September 14):

| Round | Seat A | Seat B | Accepted | Rejected | Needs a human |
|---|---|---|---|---|---|
| 1 | 5/5 (0) | 3/5 (1) | 1 | 0 | 0 |
| 2 | 5/5 (0) | 5/5 (0) | 0 | 0 | 0 |

Seat A used the documented substitute after the configured reviewer failed authentication on both attempts. The accepted finding aligned the fourth phone validator in the Mac contact-review resolver.

Latest Contacts fallback review (September 14, `423496c`):

| Round | Seat A | Seat B | Accepted | Rejected | Needs a human |
|---|---|---|---|---|---|
| 1 | 4/5 (3) | 5/5 (0) | 1 | 2 | 0 |
| 2 | 5/5 (0) | 5/5 (0) | 0 | 0 | 0 |

The accepted item added coverage that a valid zero-match native response still runs the AddressBook exact phone/email scan. Both final reviewers agreed with the disposition of the other two suggestions; no open findings remain in this follow-up.

## Needs a human

- Maintainer review of the new actions and Mac setup flow.
- Approve the first-time fork workflow so hosted CI can run.
- Contributor device retest on macOS 26.3 remains useful; the nil-predicate failure is covered by synthetic regression, not a newly observed live native error.
- Native deletion currently requires English Messages labels and an unlocked Mac with Messages open. Other macOS releases/locales, group messages, and attachment-only native deletion still need device coverage. Unsupported layouts fail closed.
- Account placement follows the Mac's Contacts configuration. Deletion is not Undo Send; Messages in iCloud may sync it across the user's devices.

factory: rounds=2 evidence=A,B,D


ianswope:
I filed #78 asking whether this was wanted and how it would work, so I read this properly rather than just reacting to the diff. Short version: the mechanism is better than what I sketched in the issue, and the thing it is missing is the thing I can actually help with.

**What I think you got right, specifically.**

My worry in #78 was balloon selection — a transcript search by text is how you delete the wrong message. You avoid that entirely. Focusing `messageBodyField`, then *verifying the previous selection cleared* before opening the link ("so an unsupported link cannot delete a stale one"), then letting `sms://open?message-guid=` make Messages select the exact balloon, with text only as a cross-check, is a stronger design than matching text yourself. And `outer.count == 1` plus `texts == [expected]` means an ambiguous tree fails closed rather than guessing.

`imsg-delete` also keeps the discipline `imsg-read` paid for — the comment that a successful accessibility action is not proof, then polling for the row to leave and returning `unverified` when it does not. That is the right referee.

Things I checked rather than assumed:

- Text reaches the helper on stdin, never argv, on both machines.
- SQL is read-only; the tools never write `chat.db`.
- The Swift is shipped as source and run through `/usr/bin/swift`, so it stays auditable — no binary in the tree.
- `check_mutations()` uses the **non-prompting** `AXIsProcessTrusted()` and sits behind `--mutations`, so it does not fire a consent prompt nobody asked for. Given #36, that distinction matters a lot, and it would have been easy to get wrong.
- `imsg-delete` loads `imsg` via `SourceFileLoader` to reach `decode_attributed_body`. That is safe **today** only because `imsg` guards its entry point with `if __name__ == "__main__"` (line 2172). It is worth a comment: if `imsg` ever grows import-time side effects, deletion breaks in a way that will not look related.

**The gap, said plainly.** The body says the Swift helper *typechecked* on 26.5.2, and the images are *synthetic UI captures from the real QML components*. So the QML renders and the Python has unit tests, but the accessibility path — the part that actually deletes — has not been run against a live Messages. Everything downstream of `AXUIElementPerformAction` is currently unexercised: the localized action name, the first-use Recently Deleted notice, the two-button sheet, and the 5-second verification window. Your own "Needs a human" says as much; I am agreeing with you, not catching you out.

**Offer.** I can give you that device coverage. My gateway is macOS 26.3 with `chat_recoverable_message_join` empty — 0 rows against 11,760 messages — so a delete moves it 0 → 1 with no ambiguity about what caused it, and I would start in the self-thread where the result is recoverable. Say the word and I will report what happens, including the failure modes.

**Three smaller things while I am here.**

1. **Focus.** `NSWorkspace.shared.open` on the `sms:` link will pull Messages forward on the Mac. That is the same trade-off `imsg-read` documents for `--chat` ("MUST open the conversation, which pulls Messages forward — that is why `push_read` defaults to `all`"), and it is why per-thread read-push is not the default. Worth stating in `MAC-ACTIONS.md`, since someone at the Mac will see it happen.
2. **Locale.** The English dependency is doubled: `deleteAction` matches `Name:Delete…`/`Name:Delete...`, and the sheet handling matches `AXTitle == "Delete"`/`"Cancel"`/`"OK"` plus the exact string `Deleted messages are moved to Recently Deleted.`. Failing closed is the right call; I would just say "English Messages only" in the README rather than only in the PR body.
3. **Undocumented surface.** `sms://open?message-guid=` is load-bearing and not a documented Apple interface. It fails closed if it changes, which is fine, but it belongs in `SECURITY.md` next to the honest note you already wrote about the key, so a future reader knows which part is built on sand.

On scope, that is Fred's call and not mine: this bundles contact-save with message-delete, and it is local Delete rather than the Undo Send half of #78. Both are reasonable places to draw a line — I mention it only because #78 asked about both and someone reading this later should not assume the other half is covered.

ianswope:
Device coverage, as offered. macOS 26.3 gateway, Blip 2.5.0, tools staged in a temp directory rather than `~/.blip/bin` so my working bridge stayed untouched (confirmed afterwards: still the original nine files). One test card created and removed; nothing else changed.

## Grant state first, because it bounds what anyone can test

Your own non-prompting probe, run against this Mac:

```
{"accessibility":false,"contacts":3}
xcode-select: /Applications/Xcode.app/Contents/Developer
swift: /usr/bin/swift
```

So by `check_mutations()`'s own rule, **`contact_save` is ready here and `message_delete` is not.** Worth stating plainly for the thread: the deletion half cannot be exercised by anyone who has not first granted Accessibility to `sshd-keygen-wrapper` — which is the grant #36 is about. I have not granted it, so **everything below is contact-save only.** The accessibility path remains untested by anyone, and `swiftc -typecheck` in the new CI job does not change that.

## What works

- **Create.** `{"ok":true,"created":true,"id":"…:ABPerson"}`, and the card is independently visible through `contacts lookup` afterwards — not just self-reported.
- **Duplicate detection**, re-running the identical request: `{"ok":false,"code":"duplicate"}`.
- **`confirmed:false` is refused**: `{"ok":false,"code":"invalid"}` — the guard holds when run for real, not only in `test_contact_save.py`.
- **Failures create nothing.** After both failing cases below, exactly one card existed. Nothing partial, nothing orphaned.
- Local CI is green on the branch here: 499 `bun test`, all three new Python suites, every Mac tool `py_compile`s, shims `bash -n` clean.

## Two findings

**1. `(212) 555-0187` is rejected as `invalid`.**

```
{"operation":"create","confirmed":true,"handle":"(212) 555-0187", …}  →  {"ok":false,"code":"invalid"}
```

`identity()` requires `\+?[0-9]` at the *start* of the raw string, so a leading parenthesis fails before normalization ever happens. The digit count it goes on to check (`5 <= len(digits) <= 15`) would have accepted it fine. This is the most common way a North American number is written, and it is how Contacts itself displays one — so a user copying a number out of Contacts and pasting it into the save form gets a flat "invalid" with nothing indicating that the parentheses are the problem.

**2. The same number without a country code returns `unavailable`, not `duplicate`.**

```
{"operation":"create","confirmed":true,"handle":"2125550187", …}  →  {"ok":false,"code":"unavailable"}
```

with `+12125550187` already saved. The bridge's own reader resolves these to the same person — `contacts lookup 2125550187` matches the stored card — so the regional matching exists; the create path just does not reach the same conclusion.

Two things about this rather than one:

- The **code is misleading**. `unavailable` is a catch-all shared with permission and adapter failures (including the `throw new Error('unavailable')` when `nativeLookup()`'s predicate result unwraps to nothing). A user who is in fact looking at an existing contact is told the store is unavailable.
- Worth checking which branch actually produced it. Nothing was created, so this **failed safe** — but if that is an error path rather than the duplicate check returning `true`, then the thing that prevented a duplicate here was an exception, not the duplicate logic. I could not tell from the outside which it was, and the difference matters: the first case would create a duplicate the moment the predicate returns an empty-but-valid result.

Reproducing both needs only a reserved fictional number (`+1 212 555 0187`) and a Mac with Contacts authorized.

## Cleanup

Test card deleted (matched on last name `PR79Test`, verified gone), staged directory removed, `~/.blip/bin` never written to. Happy to re-run any of this, and happy to cover the deletion path too if I end up granting Accessibility — but I would rather that decision be made deliberately given #36, so I am not treating it as a formality.

Kb2uka:
@nixfred could you review this PR and approve its fork workflow when you have a chance? You may know me as @StudioXRadio on X.

These actions matter because many people moving their daily desktop work to Omarchy keep a Mac running headless as their iMessage gateway. Saving a new contact and deleting an individual message from Blip lets them do that work from Linux.

There is a limitation we need to be clear about: deletion operates the Messages interface through Accessibility. A monitor is not required, but the Mac must be awake, logged in, unlocked, and have Messages open. Screen Sharing can prepare that session. The Mac's Messages scripting dictionary has no individual-message deletion command. This does not provide deletion from a locked desktop, and it is not Undo Send. Messages comes forward during the operation.

@ianswope thanks for the detailed checks. Follow-up audit:

- Reproduced the parenthesized-number rejection and corrected the Linux, Python, and native helper validators, with failing-before/passing-after regression coverage.
- Could not reproduce the country-code-omitted number returning `unavailable` on macOS 26.5.2 at the original PR revision. With the same fictional number already saved, the native predicate returned zero matches, the fallback supplied one candidate, and the regional matcher correctly returned `duplicate`. Could you share the exact staged revision/helper hashes and the Mac's `AppleLocale` setting so we can narrow down the difference? I have not claimed a fix for a failure I could not reproduce.
- Found a separate locked-desktop failure: Messages exposed a repeating accessibility element, exhausting the traversal limit and producing misleading “unsupported” wording. The traversal now handles cycles while retaining its size/depth limits, allowing the existing unlock/open-Messages guidance to surface.
- Added explicit headless-Mac guidance, the English-only deletion limitation, the undocumented `sms://open?message-guid=` dependency, and the import-safety maintenance comment.

For device coverage, deletion was exercised against live Messages on macOS 26.5.2 during the original contribution. I also deleted a message through the installed PR today after unlocking the Mac; the first unlocked attempt requested that Messages stay open, and the second succeeded. Broader version/layout coverage is still useful.

Verification: 500 TypeScript tests and 102 Python tests passed on Linux; 42 Python tests, the Swift traversal regression, and Swift typecheck passed on macOS. Native contact creation/readback and exact/local/parenthesized duplicate checks passed, with the synthetic card removed afterward. Hosted CI still requires maintainer approval.

ianswope:
Thanks — the parenthesized fix verifies here, and I think I can explain the
non-reproduction without either of us running anything else.

## The details you asked for

Staged revision was **`e17d1327`** (the only commit on the PR at the time),
copied out of the branch rather than edited. sha256, first 16:

```
3c31b93d1bd160dd  bridge/mac/contact-save
b68702651bb499c4  bridge/mac/contact-save.js
fc72b48900f4a4b7  bridge/mac/contacts
e28f6f03a21a25d2  bridge/mac/blip-dispatch
```

Gateway is **macOS 26.3** — yours is 26.5.2, and that gap turns out to
matter less than it looks. `AppleLocale` is `en_US`, `AppleLanguages` is
`("en-US")`, so locale is not the variable.

## Your fix for the parenthesized number

Verified at `bf14b3ce`: 500 `bun test`, and the Mac-side Python suites run
103 with 1 skipped (your 102, plus the darwin-gated Swift test that skips on
Linux). `[0-9]` → `[0-9(]` is the right minimal change — `(((((` still fails
on the digit count, so nothing loosened except the leading paren. Applying it
to all four validators rather than just the one I hit is the correct call.

## Why #2 did not reproduce for you

I do not think this is a version difference. `contact-save.js:169`:

```js
const duplicate = adapter.contains(draft);
return JSON.stringify({ok:true, duplicate:duplicate,
  phones: duplicate ? [] : adapter.phoneCandidates(draft.phone)});
```

Both halves sit inside the one `try`, and `contains()` runs first. It calls
the CN predicate, which throws at `contact-save.js:79` if the result unwraps
to nil:

```js
if (!ObjC.unwrap(found)) throw new Error('unavailable');
```

That throw takes down the whole inspect response, so `check_duplicates`
returns at its first gate (`contact-save:113`, `response.get("ok") is not
True`) — and **`phoneCandidates` never runs, so the regional matcher is
never reached.** `unavailable`, exactly what I saw, and nothing created.

The reason that is worth more than a version note: `contains()`'s
AddressBook fallback compares `saveKey()` output, which is raw digits with no
regional handling. A stored `+12125550187` keys to `phone:12125550187`; the
query `2125550187` keys to `phone:2125550187`. Those cannot match. So
`contains()` structurally cannot recognise a country-code variant — **all of
the regional matching lives in the path that a `contains()` failure
destroys.** On your Mac the predicate answered and the fallback carried it;
on mine the predicate returned nil and the regional check was unreachable.
Same code, different luck.

That also answers the question I left open last time: the thing that
prevented a duplicate here was the exception, not the duplicate logic. It
failed safe, but for the wrong reason, and the failure mode is one bad CN
answer away on any version.

A shape that would fix it: let the two halves fail independently — gather
`phones` even when `contains()` could not answer, and report `unavailable`
only if neither the exact check nor the candidate list could be produced. A
CN predicate that cannot answer is "CN could not answer", not "the store is
unavailable".

I would rather not send you chasing 26.3 specifically until that coupling is
addressed, because with it in place the symptom is reachable on any version
and the repro is "make the predicate return nil", not "run 26.3".

## One gap left over

The phone rule now lives in four files — `contact-review.ts`,
`bridge/mac/contact-save`, `bridge/mac/contact-save.js`, `bridge/mac/contacts`
— and each has its own behavioural test, but nothing compares them. Fixing
one and missing the other three would stay green. That is the same shape as
the group-chat rule in #75, which ended up with a drift test that reads the
sites and asserts they still agree; worth the same treatment here, since this
round needed all four edited at once to stay consistent.

## The traversal find

The repeating-element cycle is a good catch and not one I would have reached
from outside. Deduplicating on `kCFTypeSetCallBacks` (so real `CFEqual`
rather than pointer identity) with the bound moved onto distinct nodes looks
right, and extracting the shipped traversal and driving it with fixture
children is a nice way to cover it without a live app.

## One wording point, for @nixfred rather than for me

The PR body at `e17d1327` described the Swift helper as typechecked, labelled
the screenshots synthetic UI captures, and asked for device coverage in its
own "Needs a human" section — which is why I said the accessibility path was
unrun. The note above now says deletion was exercised against live Messages
during the original contribution. Those may both be true of different work,
but since real-hardware coverage of the delete path is the crux of this PR,
it seems worth stating plainly which parts have been run on a device, on
which macOS version, and when. I have still only exercised the contact-save
half, and the Accessibility grant remains ungiven here.

nixfred:
Thanks @Kb2uka, and thanks @ianswope for the device testing and the design write-up in #78. Selecting a message by its original row ID and GUID and then confirming it's gone from chat.db is the careful approach I'd want. I've approved the CI run for your fork.

Before reviewing, please split this into two PRs:

1. **Delete one message.** `imsg-delete`, `MessageDelete.qml`, the Swift selector, and the `blip-check --mutations` probe. It's recoverable (Recently Deleted, 30 days) and matches #78's design. Undo Send stays out.
2. **Save a new contact.** `contact-save`, `ContactSave.qml`, and the duplicate checks.

Why: each PR adds a new tool to what the confined ssh key can run in `blip-dispatch`, and each needs a different permission on the Mac (Accessibility for delete, Contacts write for save). Up to now Blip has only read the Mac, so I want to review each of those changes on its own. At 2,000 lines together, that isn't practical.

Two things to cover in the delete PR: what the user sees when Messages is locked, closed, or in a language other than English (it should refuse, not guess); and `blip-check` must not fire the Accessibility prompt unless it's run with `--mutations`, because an ignored prompt is recorded as a denial (#36).

Kb2uka:
Superseded by #90 for message deletion and #91 for contact saving, as requested. Both are separate draft PRs against main, with passing Linux and macOS checks and no current merge conflicts. The new workflows need maintainer approval to run.

PR #80, opened 2026-09-14
time: stamps cross the bridge as UTC; local time is a display concern

zachwilke:
Split out of #47 (first of five, as requested). Rebased onto current `main` (2.5.0).

## What

Stamps cross the bridge as ISO-8601 UTC (`2026-09-07T18:33:12Z`). Local time is a display concern: bubbles, day dividers, "Today"/"Yesterday" and read receipts convert to the reader's zone at render, so a day still breaks at your midnight.

## Why

Stamps arrived as the Mac's naive wall clock and were compared against the Linux clock — the same string only while both machines sat in one timezone. A Mac an hour ahead put every read mark ahead of every message (nothing unread); an hour behind and the backlog re-toasted. In the DST fall-back hour the Mac's own clock repeats, so two messages an hour apart carried the same stamp and lexical order stopped being chronological order. Every `a.ts < b.ts` in the collector already assumed that property; only UTC actually provides it.

`imsg`'s JSON emits `fmt_ts` (UTC). Plain-text CLI renders keep `fmt_ts_local` (the time a person reading `imsg recent` remembers).

Upgrading is covered from both ends, and both halves are load-bearing together: `loadState()` migrates marks a pre-UTC release wrote, and stamps are normalised at the two fetch doors (`fetchMessages`, `loadThread`, plus `fetchChats`' `last`) so a Mac still on the old bridge keeps working. Migrating the marks alone would sort every message below every mark (`" "` < `"T"`) and silently empty the badge and the toasts.

This closes the open ROADMAP item ("Timestamps are Mac-local wall-clock strings").

## How it was verified

- [x] `bun test` is green (497 tests; CI-equivalent Mac-tool tests including new `test_wire_time.py`)

invariants touched: **Every stamp inside Blip is UTC; local time is a DISPLAY concern.** Marks migrate on load; stamps normalise at the fetch doors. The suite is pinned to `TZ=UTC` (`test-setup.ts`); `timezone.test.ts` and `bridge/mac/test_wire_time.py` set their own zones and are what actually cover the conversion.

## Rebase notes

`main` moved a long way since #47. Conflict resolutions kept current-main behaviour (chat-id sanitising, pin_name, failed-send `PendingSend` fields, `XDG_CONFIG_HOME` test isolation) and applied `toUtcStamp` on `fetchChats().last`. A `selectIncomingLinks` fixture added on `main` after #47 still used a naive stamp; it is now UTC so it does not sort below a UTC watermark (`" "` < `"T"`).

Supersedes the UTC slice of #47 (`688f2cb`).


nixfred:
Merged, thanks @zachwilke! stamps now cross the bridge as UTC. It also fixed issue #86 (a send showing twice when Linux and the Mac are in different zones). Your fixes at both ends mean a Mac still on the old bridge keeps working, which made this safe to merge. It's in main as of 186b536 and deployed on my machine. I merged it together with nine other PRs, and all 526 tests pass on the combined tree.

PR #81, opened 2026-09-14
gifs: an animated attachment arrives animated, and renders animated

zachwilke:
Split out of #47 (second of five, as requested). Rebased onto current `main` (2.5.0). Independent of the UTC PR.

## What

An animated GIF arrives as its original bytes and renders through `AnimatedImage`, so it actually moves. Stills stay on `Image` (the one that applies `autoTransform`, the EXIF fall-back).

## Why

A GIF was flattened twice. `fetch.ts` treats any `image/*` under the auto-fetch cap as a preview candidate and asks the Mac to resample it with `sips`, which collapses an animation to one frame — a 1.4 MB GIF reached Linux as a 198 KB JPEG. And a QML `Image` paints one frame whatever you hand it.

Animated mimes now skip the resample and land in the `orig` cache slot a click already shares. Sizing is computed once on a wrapping `Item` so the two renderers cannot disagree about the delegate's width (a delegate that outgrows the panel takes every right-aligned element off-screen). Only the active renderer takes a source, or every photo would decode twice. `AnimatedImage` honours `sourceSize` (measured), so the decode ceiling still applies. `imageMetrics` also learned the GIF header — it reported 0×0 before.

## How it was verified

- [x] `bun test` is green (includes new GIF header / `isAnimatedMime` / QML renderer invariants)

invariants touched: **Cache file extensions follow the gated MIME**; **EXIF orientation is baked at fetch time, and every `Image` sets `autoTransform: true`** (stills stay on `Image` for that reason). Animated formats skip the sips resample so they are not flattened.

## Rebase notes

`BlipView` auto-merged onto current-main's attachment chips, including the tapback-on-picture pill from #57. CHANGELOG entry is under Unreleased; the UTC changelog line that sat next to it on #47 is not in this PR.

Supersedes the GIF slice of #47 (`1feab93`).


nixfred:
Merged, thanks @zachwilke! GIFs move now. Keeping stills on Image so photo orientation still works, and loading only the active renderer, were both the right calls. It's in main as of 186b536 and deployed on my machine. I merged it together with nine other PRs, and all 526 tests pass on the combined tree.

PR #82, opened 2026-09-14
compose: a message can carry several files, not just the first one dropped

zachwilke:
Split out of #47 (third of five, as requested). Rebased onto current `main` (2.5.0). Independent of the UTC and GIF PRs.

## What

A dropped, pasted, or `/attach`ed set of files all stay on the draft. They ship one part per file, caption on the first only, capped at ten. A failing part stops the batch and leaves the rest attached.

## Why

Dropping five photos attached one and discarded four in silence: the handler read `urls[0]` and the draft was a single path. A drop that loses most of what you dropped, with no message, is worse than one that refuses.

Sends go through the existing `send-file.ts` path (unchanged: still one file per invocation, still stdin, still refusing groups whose guid is unknown). The service is captured when the batch starts, never re-read from `root.active` per part, so switching threads mid-send cannot push a later part onto a different service (war room #2).

Chips are one per row, for the implicit-width reason the received ones are.

`main` already uses `drafts` for the per-conversation **text** draft map, so the attachment list is `attachDrafts` / `attachCount` / `addAttachment(s)` / `removeAttachment` / `clearAttachments`.

## How it was verified

- [x] `bun test` is green (new QML invariants for drop-all-files, the cap, caption-on-first, one-chip-per-row, service capture)

invariants touched: **Never send to a handle for a group** (still `--chat-id` / `--to` via send-file.ts); **Message text never rides argv** (caption still `--caption-stdin`); **Never let one delegate's implicit width exceed the panel** (chips one per row).

## Rebase notes

`pumpFileSend` sits next to current-main's richer `copyProc` (clipboard feedback). The QML tests that used to slice to `Process { id: copyProc }` now slice to `function copyText`, because `copyProc` is no longer a one-liner.

Supersedes the multi-file slice of #47 (`20f8bc1`).


nixfred:
Merged, thanks @zachwilke! every dropped file goes out, with the caption on the first one. Fixing the chat and service when the batch starts closes the switch-threads-mid-send hole. It's in main as of 186b536 and deployed on my machine. I merged it together with nine other PRs, and all 526 tests pass on the combined tree.

PR #83, opened 2026-09-14
perf: stop paying a startup toll; hold the Mac channel open

zachwilke:
Split out of #47 (fourth of five, as requested). Rebased onto current `main` (2.5.0). **Needs a close security review.**

Four commits: drop the shim preflight, hold `imsg serve` channels open, stop scanning the address book on a miss, and the Codex P1 from #47 (path allowlist before any channel starts).

## Security review — please read this first

`blip-bridged` holds two long-lived ssh sessions to the Mac. That changes what the confined key does in practice: instead of one short `imsg` process per poll, a process with `chat.db` already open sits behind `blip-dispatch` for the life of the channel.

**What stays the same (the forced-command gate and the `ssh -n` rules):**

- The dedicated key is still enrolled as `restrict,command=blip-dispatch`. The remote command on that path is the bare string `imsg serve`. Dispatch still execs only the allowlisted tools. `serve` is a new *subcommand of `imsg`*, not a new dispatch entry, and it re-dispatches through `build_parser()` — the same parser the CLI uses, so there is no second command table.
- `SERVE_ALLOWED` is read-only: `recent`, `from`, `thread`, `search`, `contacts`, `chats`, `groups`, `analyze`. `attachment` / `avatar` stream binary and would park the channel behind a 100 MB photo; `watch` blocks; `serve` would recurse. Those stay one-shot. Covered by `bridge/mac/test_serve.py`.
- The dedicated key still gets its **own** ControlMaster socket (`cm-blip-%r@%h:%p`). Sharing the general key's master would silently bypass the forced command. Unchanged from the shim.
- There is still **no ssh preflight** on the one-shot path, and none on the channel path either. Adding one back is a bug (CLAUDE.md): a probe without `ssh -n` eats stdin and empties `--file-stdin`; a probe at all pays the 41 ms this work exists to avoid. `blip-setup`'s *wizard* ssh calls still pass `-n` (#71) — that rule is about the setup script's own stdin, not about every ssh in the project.
- Message text still rides `stdin` (a JSON field on the socket, then the channel's stdin), never argv.
- The channel is an accelerator. No socket, no `socat`, a dead daemon, an unparseable frame — `bridgeRun()` falls through to `~/bin/imsg`. A caller that passes its own `runner` never touches the socket (keeps the suite honest).

**Codex P1 from #47, fixed here:** when the dedicated key is absent/unreadable, `_spawn()` interpolates `python` and `remote_bin` into a remote shell command. The new parser used to accept those values verbatim. A hostile `bridge.conf` could inject metacharacters and run on the Mac through the general ssh path, reopening the threat the parsed shim closed (audit #7 / AGENTS.md path validation). `blip-bridged` now applies the **same regex allowlist as the shim** in `read_conf()` and again in `_spawn()`, before any channel starts. Exit 78. Covered by `bridge/linux/test_bridged_conf.py`.

**Architecture call still yours:** ROADMAP retired `blipd`. This is a transport multiplexer, not an application daemon — no state, the leader BarWidget supervises it the way it already supervises the push watcher. Happy to drop it if you would rather not add the process; the other two perf commits stand without it. `socat` is a soft dependency for the fast path only.

## What

1. The shim no longer fires a connectivity probe before every call (41 ms, ~31%). ssh reports transport failure as 255; the shim translates to 69. `hide_spam` / `hide_unknown` flags from current `main` are kept.
2. `imsg serve` + `blip-bridged` (two pooled channels, unix socket in `XDG_RUNTIME_DIR`).
3. Contact name misses bucket by last-seven digits instead of scanning the book. Caches invalidate on what they derive from, not `chat.db` mtime. `_forget_contacts()` also clears `_SHORT_NAMES` (added on `main` after #47).

Measured (warm, one Mac): conversation open 223→83 ms, poll 160→39 ms, deep poll 707→342 ms, bridge call 132→14 ms.

## How it was verified

- [x] `bun test` is green; `test_serve.py`, `test_bridged_conf.py`, `test_phone_match.py`, `imsg_test.py`, `test_cluster.py`, `bash -n`

invariants touched: **There is NO ssh preflight in the shim**; **`bridge.conf` is data, never sourced** (now also `blip-bridged`); **The persistent channel is an accelerator; the one-shot path is the contract**; **A name lookup that misses must not scan the address book**; **Message text never rides argv**.

Supersedes the performance slice of #47 (`0727e91`, `5b8b01e`, `934ee1f`) plus the Codex P1 fix.


nixfred (changes requested):
Thanks @zachwilke, this split is exactly what I asked for, and the numbers are great: 223 to 83 ms to open a conversation and 160 to 39 ms per poll. Keeping the one-shot path as the only thing that has to work, and making `SERVE_ALLOWED` read-only, are the right calls. I'm OK with adding the daemon.

One change before it can merge, in how the socket directory is found:

- `runtime_dir()` in `blip-bridged` and `BRIDGE_SOCK` in `collector.ts` both fall back to `/tmp/blip-<uid>` when `XDG_RUNTIME_DIR` is unset. `os.makedirs(..., mode=0o700, exist_ok=True)` accepts a directory someone else already created, and the client only checks `existsSync(BRIDGE_SOCK)`. Any other local user who creates `/tmp/blip-<uid>` first owns the path. They can plant a socket that receives `argv` and `stdin` (`search --stdin` sends message text) and returns made-up JSON that Blip then shows as your messages.
- Suggested fix: on both sides, `lstat` the directory and the socket, and use them only if the directory is a real directory (not a symlink) owned by `getuid()` with mode `0700`, and the socket is owned by `getuid()`. Otherwise go straight to the one-shot path. Simplest: don't use the fast path at all when `XDG_RUNTIME_DIR` is unset. A test for the wrong-owner and symlink cases would lock it in.

It's low risk on Omarchy, where `XDG_RUNTIME_DIR` is always set, but Blip runs on other machines too. The rest of the PR tested clean on top of current main (bun and all Mac/Linux Python suites green).

nixfred:
Checking in, @zachwilke — this has been sitting since the 15th and I do not want it to rot, because the idea is good.

Two things still block it: the CHANGES_REQUESTED review about ownership of the socket directory, which has not been answered, and the branch conflicts with main. Main has moved a lot since you opened it: bin_dir= now moves the shims (#100), per-conversation spawns route through a source-id seam (#105), and imsg-read addresses groups through Messages' deep link (#106). That last one touches the channel you are holding open.

If you still want it, rebase and answer the ownership point and I will review it properly the same day. If you would rather drop it, say so and I will close it with thanks: your UTC work (#80) and the multi-file and GIF fixes are already shipping and that is real.

Issue #69, opened 2026-09-11
Outbound tapbacks do not need SIP off — the Accessibility line `imsg-read` already crossed reaches them. In scope?


Fileri:
### The idea

README ("What it can't do"), ROADMAP ("Not possible") and CLAUDE.md all say outbound tapbacks need BlueBubbles-style code injection and therefore SIP off. That was true of that approach. It is not true of the approach Blip already uses for *Mark All as Read*: System Events on Messages.app, under the Accessibility grant `imsg-read` holds today, SIP on.

Every rendered transcript balloon exposes the tapback row as AX custom actions — `Heart`, `Thumbs up`, `Thumbs down`, `Ha ha!`, `Exclamation mark`, `Question mark`, the user's recent emoji, and `Add Emoji as Tapback` (the picker, with its search field). Performing one puts a real tapback row in `chat.db` (`associated_message_type` 2000–2005 to add, 3000–3005 to remove; 2006 with `associated_message_emoji` for a custom emoji), which the collector already renders as a pill. Beeper's `platform-imessage` (MIT) drives Messages the same way on Tahoe with SIP on.

I have this working against my own Mac as a local patch, so this is a report, not a guess. Measured, with the caveat that most of it is one day of deliberate testing rather than weeks of use:

- Built on 2026-09-05 and exercised hard that day in my own self-thread: 160 reactions confirmed, 17 failed. Since then a handful of real reactions in real conversations; two failed on 2026-09-10, both on a sticker (a rendering case the matcher did not know — fixed locally).
- Every success is confirmed by the row appearing in `chat.db`, not by what the screen shows (the `report()` idea from `imsg-read`).
- Wall time from the Linux keypress to the confirmed row, over the whole log: median 3 s, p90 9 s, worst 33 s. The tail is the code's history, not its present. Two fixes on the first two days each removed a class of slow runs, and both are measured before and after: the transcript scan went from two full `entireContents()` passes per reaction to one bottom-up pass with early exit (cold ~13 s → 3.8 s, warm 2.6–3.0 s; every 15–16 s run in the log predates it), and the custom-emoji path stopped scanning the whole window with the emoji catalogue open (19–33 s → ~6 s; the log shows 33 s at 12:10 and 6 s at 12:20 the same day, then 4–5 s). A recently used emoji is a direct AX action on the balloon and takes 2–2.5 s. Since those two fixes nothing has taken longer than 6 s — a small sample, admittedly.
[...]

PR #84, opened 2026-09-14
docs: "tapbacks out is impossible" is half wrong on macOS 26

zachwilke:
Split out of #47 (docs slice only). Rebases `586a4de` onto current `main`.

## What this is

The old "Things that are not possible" note said outbound tapbacks, edits, and typing indicators were out because they need SIP-off code injection. That is **half wrong** on macOS 26: Messages now has real menu items (`Edit ▸ Tapback Message…`, `Edit ▸ Reply to Message…`, `Edit ▸ Edit Last Message…`, plus `Send Later…`). A menu item is scriptable — that is how `imsg-read` overturned the old "marking read is impossible" note. Typing indicators still have no menu item.

What is **not** proved, and what anyone picking this up has to establish first: all four items read `enabled=false` from the background (evidence of nothing — AppKit validates menus against the active app). Feasibility has to be tested with Messages frontmost. Then the real obstacles: the item acts on the selected message; selecting an arbitrary bubble from Linux is unsolved; the "…" suggests a picker; Messages must come forward; a group still cannot be addressed.

Also updates the ROADMAP sending-tapbacks line to match.

## Issue #69

[#69](https://github.com/nixfred/blip/issues/69) is the live discussion (AX-on-balloons, SIP on). Nothing from that issue is in-tree; this PR only points the docs at it so the next person does not start from the stale "impossible" note.

## What this is not

No code, no CHANGELOG, no tests. Independent of the other #47 slices (#80 UTC, #81 GIFs, #82 multi-file, #83 bridge perf).


nixfred:
Merged, thanks @zachwilke! the docs no longer call outbound tapbacks settled as impossible, and point to #69. It's in main as of 186b536 and deployed on my machine. I merged it together with nine other PRs, and all 526 tests pass on the combined tree.

PR #87, opened 2026-09-15
fix: keep the app on its workspace after idle

jefehoser:
#48 restored the app window after a shell or plugin reload. Idle is a different path: screensaver and display-off remap the shell's window as a new client on whatever workspace is focused, and the restore code then saved that workspace as home. Walking away from workspace 5 with Blip on 2 left it on 5.

This change keeps a replaceable runtime compositor rule for the live title (`Blip` / `Blip (N)`), not only the unique restore title. A user `movewindowv2` is the new home. A remap or monitor churn is sent back silently (`follow = false`) and does not take focus. Closed windows stay closed; explicit launches still work.

No permanent user workspace rule. Window identity is still shell PID plus title. Only window metadata is persisted.

Validation: 488 tests pass, including the move-vs-remap decision, live-title home rule, silent return, and a QML invariant that idle remaps cannot assign `savedWorkspace = currentWorkspace`. Live: a shell restart from another workspace restored Blip on workspace 2 unfocused; a compositor move off 2 was returned there.


nixfred:
Merged, thanks @jefehoser! Blip now stays on its workspace after idle. A user move sets the new home; a remap after idle gets sent back. It's in main as of 186b536 and deployed on my machine. I merged it together with nine other PRs, and all 526 tests pass on the combined tree.

PR #88, opened 2026-09-15
The panel hangs half-faded on open: the list renders all 300 rows

ianswope:
Clicking the bar icon paused the popout mid-fade — slightly transparent,
half a second of nothing, then it snapped in.

The fade is not slow, it is **frozen**. Omarchy's `PopupCard` fades the card
in over 140 ms and the GUI thread blocked partway through, so one frame
painted at partial opacity and nothing followed.

## Measured

A `FrameAnimation` frame-gap probe, 300 conversations, one line per open:

```
441ms  476ms  484ms  496ms  580ms  627ms
```

The thread list is a `Repeater` inside a `Flickable`, so it instantiates **and
renders** every row it is handed; about ten are on screen. The popout's layer
surface is destroyed on close, which releases the scene graph, so all 300 rows
were rebuilt from scratch on every open.

## Ablated first, because the obvious culprit was the wrong one

Each row carries a `layer.enabled` mask (its own FBO) and a `MultiEffect`
pass for the circular avatar — 300 of each looked like the answer. Removing
both changed nothing: **~480 ms of gap remained**. Capping the list at 20 rows
removed the stall entirely, and dropped the thread-model assignment from
**228 ms to 19 ms**. That is the fix this PR makes properly.

## The change

The list builds only the rows near the viewport. `rowBudget` starts at 24 and
grows a batch at a time as the reader scrolls toward the end of what is built
(`growRowsForScroll`, off the Flickable's `contentY` — the same write the wheel
handler makes). `ensureRows` covers the keyboard paths that address a row by
index, End and paging. Closing resets the budget, so the next open is cheap
again.

The model is the **count**, never a slice: a Repeater handed a new array
destroys and rebuilds every delegate, which is the cost being avoided, while an
int model only appends. A side effect is that a `threads` refresh now
re-evaluates bindings instead of rebuilding 300 rows — the rebuild that the
existing `contentY` restore in `onThreadsChanged` exists to paper over.

Two traps, both guarded and both in the tests:

- an absent row's `chat` is `""`, which is also `cursorChat` when there is no
  cursor, so a placeholder would light up and claim `cursorRow`
- `scrollCursorIntoView` measures a row, so a cursor move that just **built**
  one has to wait a frame (`cursorCatchUp`)

## Verified on the live shell

- four opens after the fix: **no frame gap at all**
- growth stepping 24 → 48 → … → 300 and stopping there
- both growth paths exercised: the scroll path fires in normal use, `ensureRows`
  is the safety net when the cursor outruns it
- before/after screenshots identical — photos, initials, group composite
  avatars, blue-dot slot and previews unchanged
- `bun test` green; two new tests guard the invariant

**One caveat for a reviewer:** I drove the scrolling synthetically (no
`ydotool` on this box), so real mouse-wheel growth was not exercised directly.
The wheel handler's only action is writing `contentY`, which is the trigger
that fired — but it is worth someone scrolling a long list by hand.

New invariant in `CLAUDE.md` beside the other list and scroll rules.


nixfred:
Merged, thanks @ianswope! the list now builds only the rows near the viewport, so the panel fade doesn't freeze. Ruling out the avatar masks before landing on the row count is exactly the right order. It's in main as of 186b536 and deployed on my machine. I merged it together with nine other PRs, and all 526 tests pass on the combined tree.

PR #90, opened 2026-09-16
feat: delete individual messages through Messages

Kb2uka:
## Summary

Companion PR: #91. Both target `main` independently.

Splits the message deletion half of #79 as requested in #78. Adds only `imsg-delete` to the confined key. The confirmation preserves the original row ID, GUID, and chat; the Mac helper selects the message through Accessibility and verifies its removal with a read-only chat.db query. Undo Send is excluded.

`blip-check --mutations` checks only deletion prerequisites through the non-prompting Accessibility API. It does not require Contacts write permission. Closed, locked, non-English, and ambiguous interfaces refuse safely. Setup, recovery, focus changes, and English-only support are documented in `docs/DELETE-MESSAGE.md`.

## Proof

Fresh captures of the real QML component on a remote Linux host, with synthetic data and a stub helper. These verify rendering and are not evidence of native Mac mutation.

| Confirmation | Refusal |
|---|---|
| Confirmation | Refusal |

The refusal guidance regression failed before the correction (5 pass, 1 fail) and passed afterward. The Mac synthetic Accessibility harness exercises the shipped Swift logic and verifies closed Messages, a cyclic locked-desktop tree, a non-English action, and denied permission all refuse without invoking Delete.

Regressions also failed before and passed after fixing post-action Accessibility errors that skipped database verification, keeping a cancelled peek unread, and distinguishing a missing helper or offline bridge from an uncertain deletion. The UI was rendered again after the review fixes. Earlier live deletion evidence is recorded in #79; this split does not claim a new live message deletion.

## Gates

537 TypeScript tests and 99 Python tests passed on a remote Linux host. On macOS, 25 Python tests, including the synthetic Swift harness, and Swift typechecking passed. Python compilation, Bash syntax, ShellCheck, and diff whitespace checks passed. Both remote hosts tested the same verified Git bundle. [Hosted CI requires maintainer approval](https://github.com/nixfred/blip/actions/runs/35043482274).

## Review rounds

| Round | Seat A | Seat B | Accepted | Rejected | Needs human |
|---|---|---|---|---|---|
| 1 | 3/5 (12) | 4/5 (3) | 10 | 4 | 1 |
| 2 | 3/5 (9) | 5/5 (0) | 1 | 7 | 1 |
| 3 | 3/5 (4) | 5/5 (1) | 0 | 5 | 0 |
| 4 | 3/5 (7) | 5/5 (0) | 0 | 7 | 0 |
| 5 | 3/5 (5) | 5/5 (0) | 0 | 3 | 2 |

Stopped at the five round limit. Accepted implementation defects were fixed and covered by regressions. Seat A's final major item explicitly routes acceptance of the requested Swift and deletion capability to the maintainer; it is not a code defect. The first attempt from Seat A in round 2 returned no verdict and was rerun successfully.

Remaining: the optional prerequisite probe detects a configured developer directory rather than proving every partial or damaged Swift installation can run. Native compilation passed on the tested Mac. This limitation and the capability decision are listed below for maintainer review.

## Needs a human

Maintainer review of the deletion capability and confirmation flow, plus approval for the fork workflow. Installing this PR adds deletion to the confined key for Macs that already granted Accessibility; the upgrade consequence is documented in `docs/SECURITY.md`. No merge or deployment is included.

The prerequisite probe uses `xcode-select -p` to detect a configured developer directory. It does not prove every partial or damaged Swift installation is usable; the native compilation gates passed on the tested Mac. Clean or repaired toolchain setup still needs maintainer coverage.

Live coverage on other macOS releases, locales, groups, and attachment-only messages remains limited as recorded in #79. Unsupported native layouts fail closed.

factory: rounds=5 evidence=A,B,D


nixfred:
Same question I am asking everyone with work in flight: we publish a weekly update crediting contributors, and we tag by X handle when we can verify one. Nothing is published on your profile or on kb2uka.com. If you want to be tagged, reply with the handle; if you would rather not be, say so and we will credit you as Kb2uka.

No pressure on the PR itself, it is still a draft and CI is approved whenever you mark it ready.

nixfred:
@Kb2uka I owe you a correction, not a review.

On 2026-09-15 in #78 I said outbound deletion was in scope as recoverable `Delete…`, that Undo Send was out, and I asked you to split #79 into a delete PR and a contact-save PR. You did exactly that, on the same day, with the edge cases covered: Messages closed, Mac locked, non-English actions, denied Accessibility, and a permission check that does not trigger a prompt. Then on 2026-09-18 I wrote the opposite in that same issue without acknowledging I was reversing myself. This PR exists because you built what I asked for in between. That is on me, not on you.

Where I land, and it is final: **Local Delete is out.** It removes my own history with no undo from Blip's side, driven by a UI matcher where a missed AX target hits the wrong balloon and reports a mis-target as a failure rather than detecting it. Messages already does this itself. The convenience is not worth a destructive action riding on UI automation.

So I am closing this rather than leaving it open to rot. It is not a quality judgement: CI was green, the refusal-on-ambiguity design was right, and your instinct to make the helper refuse when it cannot identify the message safely is the instinct I want in this codebase.

Two things I would genuinely take from you:

1. **Undo Send**, once #36 is settled. Bounded to two minutes, own messages only, behind a flag, with a hard refusal past the window rather than an attempt that silently does nothing. That one asks Messages to do a thing Messages already offers, and the worst case is that it does not work, not that something is lost.
2. **#91**, the contact save, is a separate risk and still open on its own merits. It needs a rebase on current main.

Not before #36, because I do not want new code asking for new consent while the macOS 26 Automation grant is unsettled. Thank you for asking before building, and for not probing System Events to find out. That restraint is the reason the grant Blip depends on is still intact.

PR #92, opened 2026-09-16
Offer security codes beside the focused field

bbishdotdev:
## What

`otp_autofill=on` offers incoming security codes beside the focused field, using
Linux accessibility. The prompt shares Blip's fonts and colors, handles separate
digit boxes, and falls back to the top right when field bounds are unavailable.
Codes remain available for five minutes across focus changes. No browser
extension or clipboard is needed.

## Why

The existing code toast requires a copy or keyboard shortcut. This lets the user
fill from the field they are already using. The collector and Mac bridge stay
unchanged; setup covers SMS forwarding and browser accessibility.

## How it was verified

- [x] `bun test` is green (CI will check) — 539 passing after merging current `main`
- [x] Any send/receive behavior was exercised against **my own number only** — never a contact or a group. No send behavior changed; automated verification used synthetic codes.
- [x] UI change → screenshot attached

Synthetic codes filled single fields and digit groups in Brave and Chromium,
including delayed auto-advance and a leading zero. Unrelated focus stopped
insertion. Native GTK insertion, dismissal, expiry and domain matching passed.
Six Python adapter cases and all ten existing Mac Python suites pass. Python
and shell syntax checks pass; shellcheck is unavailable locally.
Chrome, Firefox and Zen have not been live-tested.

Invariants touched: five-minute code lifetime; message text stays off disk and
argv; configuration remains in `bridge.conf`; only the leader starts the helper.

Blip's prompt beneath the code fields in Brave (phone number blurred):

Blip autofill prompt beneath six verification-code fields in Brave


nixfred:
Merged as b20e8cd. Thanks @bbishdotdev. I tested it merged onto current main in an Omarchy VM, with real mouse input, synthetic codes and a synthetic sign-in page in Chromium with --force-renderer-accessibility=complete:

- one-time-code field: prompt under the field, Fill code inserted the code exactly
- six digit boxes: filled across all six with auto-advance, leading zero intact
- a field labelled Message: no prompt
- a plain field labelled Notes: manual prompt, and a click inserted the code there, which matches what SECURITY.md documents for the manual fallback
- the clipboard never held a code

The five-minute expiry I left to your unit tests. One note for anyone writing a fake bridge: codes de-duplicate by chat.db id, so a test that reuses one id only ever sees the first code. bun test 555 pass, Python adapter 6 OK. It is deployed here with autofill left off.

PR #93, opened 2026-09-16
fix: Super+M can close unrelated windows titled Blip

jefehoser:
Super+M can close an unrelated window whose title starts with “Blip.” The documented shortcut takes the first prefix match from `hyprctl clients`; if that happens to be the focused browser/editor window, it closes that window instead of opening Blip. The app IPC focus command uses the same broad lookup and can focus the wrong window too.

This limits both lookups to class `org.quickshell` and the complete app title: `Blip` or `Blip (N)`. It follows the identity check already used by `BlipWindow.qml`. The real app keeps its documented behavior: focused → close, elsewhere → focus, absent → create. The bounded map wait and compositor lookup that avoids stale plugin IPC are unchanged.

### Regression coverage

Eight tests execute the actual README shortcut and QML shell command, with synthetic client metadata and mocked compositor/IPC commands. Unrelated Codex, browser and Quickshell windows appear before the real app, including a browser titled exactly `Blip`. Coverage includes plain and unread-count titles, focus, close, and the app never mapping. No real conversations are opened or messages sent.

- `bun test`: **535 pass, 0 fail**.
- All eight new cases fail against the original implementation.
- `omarchy plugin validate .` and `git diff --check`: pass.

Existing installations need to replace the copied Super+M binding with the corrected README example; updating the plugin alone cannot change a user's Hyprland configuration. No UI layout changes or new dependencies.


nixfred:
Merged as 681a0e6. Thanks @jefehoser. Running the real README shortcut and the real QML command against mocked hyprctl, with impostors listed first, is exactly the right way to test this. I updated my own Super+M binding to the new matcher and confirmed it finds the live window by class and exact title. bun test 538 pass on current main with both merges.

PR #94, opened 2026-09-16
Make quote-and-reply discoverable through message context menus

jefehoser:
## What

Add context menus that make replying to an older message discoverable with a right-click:

| Right-click target | Actions |
| --- | --- |
| Plain message text | Quote and reply · Copy message |
| Link or link preview | Open in browser · Copy link · Share link… |

**Quote and reply** inserts a short quotation above the existing draft and returns typing focus to the composer. It does not send anything. The keyboard quotation shortcut uses the same draft-preserving behavior. Link sharing keeps the existing QR code and LocalSend options.

## Why

Replying to a specific earlier message through its context menu is a standard part of the Apple Messages experience on macOS and iOS. Blip currently lacks that discoverable workflow: right-clicking text immediately copies it, while right-clicking a link opens the share sheet.

Blip already has a quotation workaround, but it is hidden and awkward: focus the empty composer, use **Shift+Page Up / Shift+Page Down** to select an older message, then press **Ctrl+R**. There is no visible message action that tells a mouse user this is possible. This change exposes the existing capability where users expect to find it, and avoids replacing a draft they have already started.

This provides an iMessage-style interaction for **quoted replies**, not native Apple inline reply threading. Link menus deliberately contain only link actions, avoiding duplicate copy commands and an irrelevant quotation action on the link itself.

## Screenshots

Synthetic conversations only; no real messages or contacts.

### Plain message

Plain message context menu

### Link or link preview

Link context menu

## How it was verified

- [x] `bun test`: **532 pass, 0 fail**.
- [x] `qmllint MessageMenu.qml` and `git diff --check` pass.
- [x] UI screenshots inspected with synthetic data in Quickshell; checked wide/narrow menus, keyboard navigation skipping hidden actions, quotation retaining a draft and restoring composer focus, and access to the existing share sheet.
- [x] No real send/receive tests were performed; no messages were sent to any contact or group.
- [x] UI screenshots attached above.
- [x] Invariants preserved: message/draft text remains in memory, no new send path or argv/disk transport is introduced, menu focus blocks the panel key catcher, and changing conversations closes the menu. No invariant definition changed.

Deployed locally and restarted the shell: Blip reports online and healthy with no recent Blip warnings/errors. The older installed copy has four pre-existing installer/version-documentation test failures, reproduced with the patch reversed; the PR source tree passes all 532 tests.


nixfred:
Merged as ed75965. Thanks @jefehoser. I checked it in an Omarchy VM against the shipping QML and the synthetic demo data, with real mouse input: right-click on a text bubble opened Quote and reply and Copy message, Quote and reply put the quote in the composer without sending, and right-click on a link showed only the three link actions. The one thing I could not reproduce there was focus returning to the composer, because the demo surface takes no keyboard input at all, so that part rests on your check. bun test 543 pass on main. Deployed on my machine.

Issue #89, opened 2026-09-15
Desktop catch-up notifications fire on wake for messages already read on another device (selectToasts ignores m.read)


mwhuss:
### Versions
Blip 2.5.0 · Omarchy 4.0.3-1 · quickshell 0.3.1-1

#### Summary
When waking a laptop from sleep/suspend, Blip fires desktop notifications for every allowlisted message that arrived while the laptop was suspended, even if those messages have already been read on an iPhone or another Apple device (and synced to `chat.db` via iCloud).

On a laptop that was asleep for several hours or overnight, waking the machine results in a sequential cascade of up to 20 notifications (`notify-send --expire-time=8000 --wait`), even though none of these messages are unread.

#### Cause
- While the laptop is asleep, Blip is not polling, so `state.watermark` remains at the pre-sleep timestamp.
- Inbound messages received while away are read on an iPhone/Mac. iCloud syncs `is_read = 1` to `chat.db`, which `imsg` emits on each message row as `"read": true`.
- The unread badge ledger correctly honors this: `isUnread(m, mark)` in `collector.ts` tests `m.read !== true`, so the Omarchy bar badge count stays at 0.
- However, `selectToasts()` in `collector.ts` never inspects `m.read`:
  ```typescript
  for (const m of msgs) {
    if (m.from_me) continue;
    if (m.ts <= watermark) continue;
    if (open.has(chatKey(m))) continue;
    if (!allowed.has(chatKey(m)) && !allowed.has(m.handle)) continue;
    const key = toastKey(m);
    if (seen.has(key)) continue;
    seen.add(key);
    out.push({ chat: chatKey(m), name: m.name ?? m.handle, text: m.text, ts: m.ts, key });
  }
  ```
- Because `m.read` is omitted from the gate, every inbound message newer than the pre-sleep watermark from an allowlisted sender is treated as a toast candidate.
- All matching messages get enqueued into `toastQueue` in `BarWidget.qml` (capped at the 20 newest) and drained sequentially.

#### Reproduce
1. Have an allowlisted contact send a message while the laptop is suspended (or simulate by setting `state.watermark` in `~/.local/state/blip/state.json` to an earlier timestamp).
2. Read the message on an iPhone or another Apple device so `is_read = 1` in iCloud / Mac `chat.db`.
3. Wake the laptop (or let the next poll run).
4. **Observed:** The bar badge correctly shows 0 unread, but desktop toast notifications pop up for the read messages.
[...]

PR #95, opened 2026-09-16
fix: a message read on another device never toasts

ianswope:
## What

`selectToasts()` now skips messages the Apple side already reports as read, so
waking from suspend no longer fires a catch-up notification for every message
you read on your iPhone overnight. Nothing else about toast selection changes.

## Why

#89, reported by @mwhuss with the cause already located. The gate list in
`selectToasts()` covers `from_me`, the watermark, the open conversation and the
allowlist, but never `m.read` — while `isUnread()` does test it. The two
therefore disagree: the bar badge correctly stays at 0, and notifications fire
anyway for messages nobody has left unread.

It surfaces on wake because nothing polls while the machine is suspended, so
the watermark stands still. On resume every allowlisted row that arrived in the
meantime is newer than it, and `BarWidget`'s `toastQueue` drains up to 20 of
them sequentially through `notify-send --expire-time=8000 --wait`.

The gate is `m.read === true`, the same half of the **"Unread = BOTH sides
agree"** invariant the badge already honours. An older bridge omits the field
entirely; `undefined` keeps the previous behaviour of trusting the watermark
alone, so a pre-1.9.0 `imsg` still toasts exactly as it does today.

`selectCodes()` is deliberately left alone: a security code already read on the
phone is still the one you are about to type into the browser on this machine.
Say the word if you would rather it matched.

## How it was verified

- [x] `bun test` is green — **530 pass, 0 fail**
- [x] Any send/receive behavior was exercised against **my own number only** — never a contact or a group. No send path touched; this is selection logic with synthetic rows.
- [ ] UI change → screenshot attached — n/a, no QML change

Three cases added to `collector.test.ts`: a message read on another device is
dropped, a simulated overnight backlog toasts only the row that is still
unread, and a bridge that reports no read state toasts as before. The first two
fail against the current implementation (189 pass, 2 fail in `collector.test.ts`
with the one-line gate reverted); the third is a guard that passes either way by
design.

Not verified against a live wake: my gateway Mac is offline after a DFU restore
(the confined key is no longer authorised), so a real poll was not possible.
The invariant is unchanged and needed no doc edit — this makes the toast path
obey what `CLAUDE.md` already says about unread.

Invariant touched: **Unread = BOTH sides agree** (1.3.2).


ianswope:
The gateway Mac is back (the key was lost in a DFU restore, not a Blip problem), so here is the live wake check the description said was missing. Read-only: it fetches the recent window through the bridge, simulates a suspend by holding the watermark back, and runs `selectToasts()` twice — once with `read` stripped from every row, which is today's behaviour, and once as shipped. Nothing is written to `state.json` and no message text, name or handle is printed.

```
oldestRowInWindow:  2026-09-12T13:52:34Z   (400 rows, so a 48h suspend fits inside the window)
inboundSinceCutoff: 38      allowlistedOfThose: 28
readTrue: 28   readFalse: 0   readMissing: 0

simulated suspend   toasts before   toasts after
14h                 0               0
24h                 0               0
48h                 28              0
```

28 notifications on wake, every one of them for a message already read on the iPhone, with the bar badge at 0 behind them — the cascade @mwhuss describes. The 14h and 24h rows are a quiet stretch in my own window, not the gate doing work.

Two controls, so "0 after" is not just the gate suppressing everything:

- `readMissing: 0` — this bridge reports `read` on every row, so the gate is reading real Apple state rather than falling through on `undefined`.
- Taking one of those same rows, flipping it to `read: false` and re-running the gate: it still toasts.

No behaviour changed since CI went green; this only adds evidence.

nixfred:
Merged in 681a0e6's parent, cd5f856, and it closes #89. Thanks @ianswope, and thanks @mwhuss for a report that already had the cause located. The live read-only wake check you ran (28 toasts before, 0 after on a simulated 48 h suspend) is what made this an easy merge. Deployed on my machine; bun test 530 pass on current main.

PR #96, opened 2026-09-16
prefer_imessage=on keeps mixed 1:1s on iMessage

bpiland:
## Why

A 1:1 that used to be iMessage and then got an RCS inbound (someone in a mixed-platform group, Continuity falling back) currently sends on the last inbound service. AppleScript does not do the Messages GUI fallback. The Mac often writes an **SMS error 4** row — Blip shows **Not Delivered** — while the iPhone still delivers RCS/iMessage.

That is last-inbound-wins from 2.2.0 / PR #4, and it is still the right default for a never-iMessage green thread.

## What

Opt in with `prefer_imessage=on` in `~/.config/blip/bridge.conf` (parsed, never sourced; re-read every poll).

When on:

1. Failed newest iMessage to a phone still → SMS (the error 22 lock stays first).
2. Else a successful iMessage anywhere in the loaded window → iMessage.
3. Else last inbound / last successful outbound, as today.

`mergeChats` no longer lets the chat-list last bubble (RCS/SMS) overwrite a DM that the window already computed as iMessage. Groups still send `--chat-id` and ignore this.

Default is **off**.

## Tests

`bun test` — 548 pass, including:

- inbound RCS after iMessage history → iMessage only when prefer is on
- failed newest iMessage to a phone → SMS even with prefer on
- never-iMessage RCS stays RCS
- `prefer_imessage=on|YES` / missing file
- mergeChats does not stomp window iMessage with listed RCS when prefer is on


ianswope:
@bpiland — I hit the send-service bug live this evening and my case lands just outside what this PR fixes by default, so here are the measurements in case they are useful to you.

The conversation is a merged 1:1: keyed by an iCloud address, with the phone number as an alias of the same cluster. Every inbound row is iMessage — 106 of them — and my phone kept sending iMessage in the same thread the whole time. Only Blip went green. Traced before any change:

```
lastInboundService:       iMessage
sendServiceForMessages:   iMessage     <- the window already agrees with you
chatListRowService:       RCS          <- the cluster's newest row, which was Blip's own last send
threadServiceAfterMerge:  RCS
blipWouldPass:            --service RCS
```

So it is your `mergeChats` observation exactly — the chat-list last bubble overwriting a DM the window computed as iMessage — but with a loop in it: the green send becomes the cluster's newest row, which makes the list row green, which makes the next send green. It does not recover on its own.

Because the flag gates that half too (`mergeChats([windowThread], listed, {}, {})` still yields `RCS` in your test), `prefer_imessage=off` leaves this broken, and I do not think a user would find the knob from the symptom — from the outside it reads as Blip deciding to stop using iMessage.

I opened #97 with just the unconditional half: a DM the window computed as iMessage is not turned green by the list. A sweep of my live window says it is narrow — 3 of 48 conversations change, and all three have zero green inbound rows ever, so no genuinely green thread moves.

Your case — an RCS inbound genuinely arriving in a thread with iMessage history — is a real judgement call and I think the flag is right for it. Ordering is yours and Fred's call: happy to rebase #97 under this, or to close it and have you take the `mergeChats` change out from behind the flag here instead. No strong preference, they just should not both land as-is.

bpiland:
Thanks @ianswope — I agree with the split.

The `mergeChats` overwrite shouldn't sit behind a flag: if the window already computed iMessage, the list row shouldn't turn the next send green. I'll take that change out from behind the flag here, as you offered.

The other half is a 1:1 that really does have RCS inbound, but also has iMessage history. That's the judgement call, and I agree it belongs on a flag: `prefer_imessage=on` to stay iMessage, default off (last-inbound). A thread that never had iMessage stays green either way.

I'll credit #97 in the PR body and CHANGELOG (Co-authored-by on the commit if it's a close lift of your diff). I'll leave #97 for you to close if this looks right. No strong feelings if you'd rather rebase yours on top of this instead.

nixfred:
Merged as ecb00b3. Thank you @bpiland — the diagnosis is the valuable part: AppleScript does not do the GUI's fallback, so one RCS inbound turns an iMessage thread green and the Mac writes an SMS error 4 row while the iPhone delivered it fine.

Shipping it off by default is the right call, and the failed-iMessage lock running first is what makes it safe. One line I added to the merge for whoever turns it on: "a successful iMessage anywhere in the window" includes an old one, so a contact who has since moved to Android gets addressed on iMessage until a send fails and the error-22 lock takes over. That is the trade the key buys, not a defect.

Only conflict was CHANGELOG.md against three merges that landed ahead of you today; both entries kept. bun test 577 pass, plus a live collector run with the key absent: services unchanged.

PR #97, opened 2026-09-16
fix: the chat list never turns a blue DM green

ianswope:
## What

`mergeChats` no longer lets `imsg chats`'s service turn a DM green when the
message window already computed it as iMessage. Nothing else about service
selection changes: a genuinely green DM still follows the list, a thread moving
*onto* iMessage still follows the list, and groups are untouched because they
send by `--chat-id`.

## Why

Found live this evening, on my own machine, as "my replies are going out via
SMS and not iMessage."

`buildThreads` picks a DM's send service with `sendServiceForMessages` (PR #4):
the last **inbound** row, i.e. the service the other person's device actually
used. `mergeChats` then overwrote it with the chat-list row's service — and
that is the **cluster's newest row**, which includes Blip's own sends, and for
a merged 1:1 (an email iMessage row and a phone SMS row folded under one
`group_id`, per the cluster invariant in `CLAUDE.md`) the phone alias counts too.

That is self-reinforcing:

1. One send goes out green.
2. It becomes the cluster's newest row, so the list row's service is now RCS.
3. `mergeChats` overwrites the thread's iMessage with RCS.
4. `BlipView.qml:1436` appends `--service RCS` for any non-group thread whose
   service is SMS/RCS.
5. Back to 1. The conversation never recovers on its own.

The conversation that surfaced it is keyed by an iCloud address with the phone
number as an alias. Traced on live data before the fix:

```
lastInboundService:       iMessage
sendServiceForMessages:   iMessage     <- the window, PR #4's rule
chatListRowService:       RCS          <- the cluster's newest row: Blip's own last send
threadServiceAfterMerge:  RCS
blipWouldPass:            --service RCS
```

106 inbound rows in that thread, every one of them iMessage. The same thread
sent iMessage from my phone the whole time, which is what made it look like
Blip had simply decided to stop.

## Relationship to #96

@bpiland's #96 fixes this same overwrite, but behind the new `prefer_imessage`
flag — its own test asserts the default still yields `RCS`:

```js
expect(mergeChats([windowThread], listed, {}, {})[0]!.service).toBe("RCS");
expect(mergeChats([windowThread], listed, {}, {}, true)[0]!.service).toBe("iMessage");
```

I think this half wants no opt-in: no preference is involved when every inbound
row in the thread is iMessage and the only green thing in sight is Blip's own
previous send. #96's genuinely mixed case — an RCS inbound arriving in a thread
with iMessage history — is a real judgement call and the flag suits it. Happy to
rebase onto #96 in whichever order suits, or to fold this into it.

## How it was verified

- [x] `bun test` is green — **564 pass, 0 fail**
- [x] Any send/receive behavior was exercised against **my own number only** — never a contact or a group. I sent nothing while testing this: the self-thread is iMessage and would not exercise a merged cluster, so the evidence below is the resolved service, not a delivery.
- [ ] UI change → screenshot attached — n/a

The merged-1:1 test fails against current `main` (197 pass, 1 fail in
`collector.test.ts`); the other three are guards that hold either way.

A read-only sweep of the live window, before/after, asking which conversations
change and whether each change is justified by that thread's own history:

```
threads in window: 48, changed: 3
  Nancy Swope     RCS -> iMessage   lastInbound=iMessage  inboundGreen=0  inboundBlue=106
  Fred Swope      SMS -> iMessage   lastInbound=iMessage  inboundGreen=0  inboundBlue=3
  +1630XXXXXXX    SMS -> iMessage   lastInbound=iMessage  inboundGreen=0  inboundBlue=18
```

Three of 48, and all three have **zero** green inbound rows in their history —
no genuinely green thread moves. Deployed here and the shell is healthy with no
Blip log errors.

Invariant touched: **Pins and merged 1:1s use the whole conversation cluster** —
this is the send-service consequence of the same folding.


ianswope:
Confirmed by delivery, which the description above could not claim: with this branch deployed, a reply sent from Blip to that merged 1:1 went out as **iMessage**. Same conversation, same machine, no config change — `prefer_imessage` does not exist on this branch.

Before the fix that thread had produced an unbroken run of green sends from Blip while the identical thread sent iMessage from my phone. It is blue from Blip now.

bpiland:
Thanks for opening this and for the live trace — I agree with the split.

I'm going to take the unconditional `mergeChats` guard into #96, keep `prefer_imessage=on` as an opt-in for genuine mixed inbound, and credit you / this PR there.

I'll leave this PR for you to close if that looks right. Happy to wait if you'd rather rebase this on top of #96 instead.

ianswope:
That's the right split — closing this in favour of #96.

The guard is the whole of this PR, so there's nothing left to rebase. Please
take the four `mergeChats` tests with it if they're useful: the first one is
the live shape (a 1:1 keyed by an iCloud address with the phone number as an
alias, every inbound iMessage, `imsg chats` reporting the cluster as RCS off
Blip's own last send), and the other three pin the directions that must stay
open — a genuinely green DM keeps SMS, a DM may still move onto iMessage, and
a group still takes the list service because it sends by chat id.

No need to credit me, but thank you. Happy to review #96 once the guard is in.

PR #98, opened 2026-09-16
perf: spend catch-up depth on the chat that needs it, not on every poll

ianswope:
Closes the ROADMAP war-room item "One never-opened unread pins the catch-up loop".

## What

The preview window fetch stops at the watermark. Each chat whose oldest
outstanding unread sits below what the window reached gets **one bounded
`thread --chat` fetch of its own** (400 rows, doubling to 3200 only for that
conversation) and those rows join at the ledger.

## Why

The ledger has to cover every outstanding unread row so a message deleted or
read elsewhere is reconciled. That boundary is **per chat** (`unreadOldest`),
but it was collapsed into one global minimum and handed to the window fetch as
its cutoff. One never-opened dot therefore set the fetch depth for *every*
poll, for as long as it sat there: 150 → 300 → … → 8192 rows across that many
**sequential ssh calls**.

Measured through my own bridge, one poll, a single stale unread:

| stale dot | bridge calls | rows | wall clock |
|---|---|---|---|
| 1 day | 1 | 150 | 380 ms |
| 7 days | 3 | 600 | 1.07 s |
| 45 days | 6 | 4798 | **3.18 s** |

Against a 6 s poll timer when the watcher is down.

After, same 45-day dot: **3 calls / 950 rows / 1.39 s** in my busiest
conversation, **2 calls / 251 rows / 0.9 s** in a quiet shortcode — which is
where a never-opened dot actually lives. Both reach their boundary.

Catch-up rows deliberately join at the ledger and nowhere else — not the
watermark, the failure ring or the toast gates. Every one of them is older than
the watermark by construction, and a message that scrolled out of the window
months ago must not toast now.

## Two judgement calls worth review

1. **A chat the catch-up cannot verify keeps the count it had** — fetch failed,
   more rows than the ceiling, or past the four-per-poll cap. The ledger only
   grows there. A dot that is really gone survives until the conversation is
   opened (which deletes its entry outright) or a later poll reaches it;
   the other direction would silently drop a real one. The capped global walk
   already carried those same chats forward, so the direction is not new.
2. **`imsg thread` bounds by row count, not date.** A `--since` on the bridge
   would make each of these one exact call. That is a vendored-tool change, so
   I left it as a note in the code rather than reaching into `bridge/mac`.

## How it was verified

- [x] `bun test` is green — **569 pass, 0 fail** (9 new)
- [x] Any send/receive behavior was exercised against **my own number only** — never a contact or a group. Nothing here sends; the new bridge call is a read.
- [ ] UI change → screenshot attached — n/a

Beyond the unit tests, I ran the real `collect()` end to end against a
sandboxed `HOME` (real bridge, planted ledger), old code vs new, with a
45-day-old unread:

```
OLD: 3352 ms, 3154 ms    unread=6
NEW:  941 ms,  959 ms    unread=6
```

Same unread count, ~3.4× faster. The planted boundary was a *fictional* unread
with no row behind it, and the per-chat fetch reconciled it away in one targeted
call — which is the job the global walk was doing.

Invariant touched: **Unread is ledger-backed** — "Catch-up fetches cover new
arrivals and the oldest outstanding unread so deletions are reconciled" still
holds; the coverage is now per chat rather than one global window. Happy to
write that sentence into `CLAUDE.md` if you want the distinction recorded there.

## Not in conflict with #83

#83 swaps the *transport* inside `fetchMessages`/`fetchChats` (`bridgeRun`) and
does not touch `fetchMessagesAfter`'s escalation or the cutoff in `collect()`.
This changes which rows are asked for, not how the ask travels, so the two are
complementary — held open channels make the remaining calls cheaper still.


ianswope:
Added the `CLAUDE.md` note in 62e8328, since the offer in the description was easy enough to just make concrete — revert it if you would rather the invariant stayed as it was, or tell me to trim it.

It records the four things this change settles: the window covers new arrivals, each chat's boundary is spent per chat, catch-up rows join at the ledger and nowhere else (they predate the watermark, the failure ring and the toast gates by construction), and an unverifiable chat keeps its count rather than reporting an undercount. The 45-day measurement is in the note too, so the next person does not have to rediscover why the global cutoff was wrong.

nixfred:
Merged as befff00. Thank you @ianswope — this is the war-room item nobody had cracked, and the per-chat boundary is the right cut.

I re-measured it here rather than taking the numbers on trust: crafted a six-week-old boundary on a busy chat in a sandboxed HOME (so the live ledger was untouched) and ran both branches twice against the real bridge. main 1.82 s and 1.79 s, this branch 0.63 s and 0.63 s, unread 3 in all four runs. The shallow poll returns 15 threads instead of 168 now, which is just the window no longer being dragged back; the panel list comes from the deep fetch.

Both judgement calls accepted as you framed them. Keeping the count on an unverifiable chat is the safe direction and matches what the capped global walk already did. And you were right to leave `bridge/mac` alone: a `--since` on `imsg thread` would turn each of these into one exact call, and it is now on the roadmap as its own change. bun test 569 pass.

PR #99, opened 2026-09-17
docs: names rank the owner's book, then the largest source

ianswope:
`b5cafc2` changed how a tie inside a source rank is settled — the largest
AddressBook source now wins outright (`_record_count` over `ZABCDRECORD`,
path as the tie-break) instead of the most common spelling, because several
small stale stores could outvote the primary account (claude-on-mac #4/#5).

The contacts invariant in `CLAUDE.md` still described the old tie-break, and
still said `_ab_sources()` returns paths in rank order; it sorts rank, then
size, then path now, which is also the order `cmd_avatar` walks for "first
source with a photo wins". This updates the note to match, including the
unreadable-Accounts-db case: everything lands in one rank, so size alone
decides rather than "the old behaviour".

Docs only — no code touched. `bun test`: 560 pass, 0 fail.


nixfred:
Merged as f490bab. Thank you @ianswope — this was our own note going stale the same day the code changed (b5cafc2), and you caught it before anyone read the wrong rule.

Checked every claim against the code it describes: `bridge/mac/imsg` sorts `(_rank_of, -_record_count, path)`, `cmd_avatar` walks that same order and takes the first photo, and an unreadable Accounts db returns an empty rank map so every source lands in rank 0 and size alone decides. All three match your wording. bun test 560 pass, 0 fail.

PR #100, opened 2026-09-17
bin_dir= in bridge.conf: let the Linux shims live outside ~/bin

cw228:
## What

A new `bridge.conf` key, `bin_dir=`, sets where the Linux-side shims (`imsg`, `imsg-send`, `imsg-read`, `contacts`) are installed and looked up. If it isn't set, nothing changes: they stay in `~/bin`.

```
bin_dir=~/.local/bin
```

## Why

`~/bin` was hard-coded in about 15 places across the TypeScript spawners, `BarWidget.qml`/`BlipView.qml` and `blip-setup`. If you keep executables in `~/.local/bin` (the XDG location), you had to keep a separate `~/bin` just for Blip.

## How

- **`bin-dir.ts`** has `parseBinDir(conf, home)`, a pure function that is built to `BinDir.mjs` so QML and TypeScript use the same parser. The file is parsed, never sourced. `~`, `~/…`, `$HOME/…` and `${HOME}/…` are expanded on this machine. Any value that doesn't end up as a plain absolute path (relative paths, `..`, shell characters) falls back to `~/bin`.
- **`shim-path.ts`** has `shimPath(tool, home?)`. Every TS spawn now goes through it: collector (recent/chats/groups/imsg-read push), thread, fetch, search, avatar, send-file, contact-search and contact-review.
- **QML**: `BarWidget` exposes `binDir` from the `bridgeConf` FileView it already has. The `imsg watch` process waits for that file's first load so it never starts from the wrong directory. `BlipView`'s send uses `hostWidget.binDir`.
- **`blip-setup`** reads `bin_dir=` (or `BLIP_BIN_DIR`, which is written back to the conf), expands it by the same rules, installs the shims there and runs the smoke test through that path.
- Docs: README, CLAUDE.md, PRIVACY.md and CHANGELOG.

## Tests

- `bin-dir.test.ts`: expansion, rejection of unsafe values, commented-out keys, `shimPath` reading a real conf, `BinDir.mjs` matching the TS source, and a guard that fails if a literal `bin/imsg` or `bin/contacts` shows up again in a spawner.
- `bun test`: 484 pass. `shellcheck -S warning` is clean on the CI's script list.
- Tried on a live Omarchy install: set `bin_dir=~/.local/bin`, moved the shims there, deleted `~/bin`, restarted the shell. `qs ipc call nixfred.blip status` reports `online=true … watch=true`, and the `imsg watch` channel comes up. The send path (`imsg-send`) wasn't exercised on the live install because that would send a real iMessage.


nixfred:
Unrelated to the review: we post a weekly public update crediting everyone whose work landed that week, and we tag people by their X handle when we can verify one. I could not find anything published for you. If you would like to be tagged, reply with the handle (or say you would rather not be, which is a perfectly good answer and we will just use your GitHub name).

On the PR itself: it conflicts with main now, and it touches 19 files. Rebase on current main when you get a chance and I will give it a proper read.

nixfred:
Thank you, this is careful work and the design decisions are the right ones. One pure `parseBinDir` built to `BinDir.mjs` so QML and TypeScript cannot drift apart is much better than two parsers, and parsing the file rather than sourcing it, then requiring the expanded value to match `^/[A-Za-z0-9._/-]+$` with no `..`, means a bad `bin_dir=` falls back to `~/bin` instead of becoming a way to run something unexpected.

Merged against current main, which had moved 73 commits ahead of your branch. The conflicts were all additive, so both sides were kept: `BarWidget.qml` (main's `otpAutofill` next to your `binDir`, in the conf handler and in `onLoadFailed`), plus the CHANGELOG and CLAUDE.md.

One change beyond your work, so you know what is in your commit: `collector.ts:1354` was still spawning `${HOME}/bin/imsg` directly for the catch-up thread fetch. That line landed on main in 0b648cd after you cut your branch, so there was no way for you to have routed it. It now calls `shimPath("imsg")` like the rest. Your own guard test, "no spawner hard-codes ~/bin", failed on the merged tree until that one line changed and passes now. That test doing exactly its job on a spawner written after the PR is the strongest argument for having written it.

Suite on the merged tree: 584 pass, 0 fail.

PR #101, opened 2026-09-17
fix: keep unread badge and thread snapshot consistent

damonjanis:
An optimistic read changes `threads` without updating the cached `threadsJson`. If suppression expires before the collector confirms the read, a repeated collector snapshot can match that stale cache: the badge updates, but the thread model keeps the read dot cleared.

Bind the comparison to the current model and derive the badge from the model actually assigned. Identical polls still avoid model replacement, and unread-message counting and optimistic suppression retain their existing behavior.

Validation:
- 560 Bun tests pass on current main plus this change.
- Four headless QtTest scenarios exercise the shipping QML bindings/functions with synthetic data: expired suppression, mark-all followed by an identical poll, in-flight suppression/new activity, and unchanged-model identity.
- The first two regressions fail against unmodified main and pass with this fix.
- The QML regression runs in CI. No layout change, real conversations, or Mac actions are involved.

Prepared with AI assistance; the fix was ported and tested independently against current upstream.


nixfred:
Merged as f7c8f29. Thank you @damonjanis — a cache that does not move with the model it describes is a good catch, and deriving it is the fix rather than remembering to update it in two places.

I ran your harness both ways instead of trusting the claim: 6 pass on the branch, and against current main's BarWidget.qml (the path argument is a nice touch) 2 fail, the expired-suppression case and mark-all followed by an identical poll. That is the first real QML regression test in this repo, and it exercises the shipping bindings rather than a copy, which is what makes it worth the CI minutes.

One note for whoever touches it next, not a change request: the harness slices functions out of BarWidget.qml by text anchors, so renaming one of those functions breaks the test rather than the code. Recorded in the merge. bun test 569 pass.

PR #102, opened 2026-09-17
fix: acknowledge read pushes only after verified completion

damonjanis:
`imsg-read --all` currently exits successfully when the unread count falls only partway (for example, 3 → 2). It also reports success if chat.db verification is unavailable. These outcomes make an incomplete or unverified push look successful in the collector's log.

Wait for zero within the existing bounded settling period, allowing partial updates and transient database read failures to settle. Return `EX_TEMPFAIL` unless the final scoped count is zero. A missing menu item can still succeed when the database confirms the conversation is read.

Validation:
- Nine synthetic tests cover partial progress, unavailable verification, zero counts, missing menu items, bounded/scoped settling, and the `--all` CLI path. Regression failures were confirmed against unmodified main.
- All 96 Mac bridge tests and 560 Bun tests pass.
- The new tests run in CI and never access Messages or a real chat database.

This only fixes acknowledgement and settling. It does not change the unread SQL, add retries, or change the read-push policy. A new unread arriving during the settling window conservatively produces failure. Fresh live macOS UI validation was not performed for this isolated patch.

Prepared with AI assistance; the fix was ported and tested independently against current upstream.


nixfred:
Merged as 134e6a0. Thank you @damonjanis — "a decreasing count is progress, not completion" is exactly right, and treating an unreadable chat.db as failure rather than success is the half of it I would have missed.

Checked the blast radius before taking it: `pushRead()` is fire-and-forget through `pushReadCommand`, which only appends `exit=<rc>` and the status line to push-read.log, so a stricter exit cannot turn into repeated Messages activation, and the per-thread path still only fires on a real unread-to-read transition. The cost is the one you named in the PR: a message landing inside the settle reports failure for a push that cleared what it could, and a non-zero settle now always spends the full window instead of returning early. Both are the right trade for an honest log.

Like you, I did not validate live: `--all` against the real gateway marks every conversation read on the account and the phone. The new tests plus the 12 bridge suites cover it, and the next real mark-all is the live check. bun test 569 pass, 12 bridge suites OK.

PR #103, opened 2026-09-18
fix: make read sync durable and add verified conversation actions

damonjanis:
Failed read pushes can leave Blip and Messages disagreeing after the local dot has cleared. This adds complete metadata-only reconciliation and a durable action worker, plus non-destructive conversation controls that use that worker.

The collector remains the only writer of state.json. Mac actions run through a flock-protected mailbox process, so slow SSH/menu calls do not block message polling. Job and intent ids prevent stale or repeated acknowledgements from clearing a newer gesture. Failures retain bounded retry state; a later per-thread gesture supersedes an unfinished global read.

Mark All Read captures the last snapshot's inbound row and timestamp boundaries. New activity cancels the pending global operation, including same-second arrivals. The Mac checks again before clicking and after waking Messages. Pin, alert and read requests are independent; completion refreshes their metadata.

User-facing changes:
- Right-click a DM to Pin/Unpin, Mark as Read/Unread, or Hide/Show Alerts. Hide Alerts suppresses Blip notifications too.
- Mac read-state reconciliation includes old unread conversations, merged phone/email aliases and reaction activity. Group unread overrides remain local.
- Badge/header totals count unread conversations, matching dotted rows. This is an intentional behavior change beyond #101's consistency fix.
- Existing Mark All Read controls, Review contact, message/link menus, Send Later, prefer_imessage and bin_dir are preserved.
- **Delete is excluded** until the selected Mac conversation can be verified before a destructive click.

Validation on this branch:
- 623 Bun tests; 120 Python tests; five headless Qt scenarios pass.
- New regressions cover slow-worker polling, restart/retry, stale completion, once-only acknowledgement, superseding gestures, same-second inbound arrivals, independent pin/alert intents, custom shim directories and local alert suppression.
- Generated QML module matches its TypeScript build. Python compile, QML parsing, bash syntax and ShellCheck pass.
- Synthetic menu screenshot inspected. No real conversations were opened, modified or messaged during this validation.

Requires Linux flock (util-linux), the new TypeScript worker/module, and both updated Mac tools with their sibling read_state.py. The default read-push policy remains `all`; per-thread pushing remains opt-in. Menu automation still depends on Messages selection and cannot atomically combine a database check with a click. Fresh live-macOS/iCloud acceptance is not claimed; those limits and release checks are detailed in [the readiness review](https://github.com/damonjanis/blip/blob/fix/reliable-conversation-actions/docs/READINESS-REVIEW.md).

Based on main after #101 and #102 merged. AI-assisted implementation and review; tests and limitations are documented above.

CI status: the initial [run](https://github.com/nixfred/blip/actions/runs/35347424599) failed before tests started because downloading Bun returned HTTP 504 on all setup attempts. Local validation above passed. GitHub refuses an author-triggered rerun without repository admin rights, so a maintainer rerun is needed.


nixfred:
Status on this one, plus two things that need you.

**Your CI failure is not yours.** The run died in `setup-bun`: GitHub's release download returned 504 three times, so bun never installed and the job never ran a test. I have re-triggered the failed job. Nothing in this PR caused it.

**It now conflicts with main.** I merged #106 (@jondkinney) a few minutes ago, which takes `imsg-read --chat` beyond DMs: it addresses a group through Messages' own `imessage:open?groupid=` deep link, serialises selection and menu clicks behind an owner-only lock, refuses to click when Messages has no accessible window, and restores the previously frontmost app. I verified the deep link live on a macOS 27 Mac before merging, because it overturns our own "a group cannot be addressed at all" note.

That lands in the same places you are working: 4 conflicts in `bridge/mac/imsg-read`, plus one each in `collector.ts`, `CLAUDE.md` and `README.md` and two in `collector.test.ts`. They are not huge, but they are in the logic both of you are changing, so I would rather you resolve them than have me guess which behaviour you intended to keep. A rebase on current main, with your DM-only assumption re-examined against group addressing now being real, and I will give the whole thing a proper read.

**One question before I merge it, whenever it is ready.** You call out that badge and header totals change to counting unread *conversations* rather than messages, matching the dotted rows, and that this is intentional and goes beyond #101. That is a user-visible policy change rather than a bug fix, so it is Fred's call and not mine. I will put it to him and report back here either way. Everything else in the description reads like work I want.

Excluding Delete until the selected conversation can be verified before a destructive click is the right instinct, and worth saying out loud.

nixfred:
Answer on the badge question, from Fred: **count unread conversations**, as you proposed. It matches the dotted rows and it matches what the iPhone badge counts, so the two stop disagreeing. Keep that part in.

Still waiting on the rebase against #106 before I can review the rest.

ianswope:
Live corroboration for one of this PR's premises, from trying to break
something else on a gateway Mac with `push_read=thread`.

Your CLAUDE.md diff says the count has to allow that "an explicit unread may
sit below `last_read_message_timestamp`". That is real, and it is currently a
silent dead end on `main`. What I did and what happened:

1. Marked a conversation unread **from the iPhone**. It synced: the Mac's
   chat.db shows `is_read=0` on one inbound row of that chat.
2. Blip kept showing it read — correctly, per "Apple-side `is_read`=0 AND
   newer than the local mark". The row predates the local mark, so it loses.
3. Reading it in Blip therefore cleared nothing, `clearedUnread` stayed false,
   `pushReadArgs` returned null, and no push ran at all. Empty
   `push-read.log`, no error anywhere.
4. Had it pushed, it would have been a no-op regardless. `UNREAD_SQL`'s
   `m.date > c.last_read_message_timestamp` does not see the row, because
   marking unread on the phone flips `is_read` without rolling the chat's read
   cursor back.

Measured on that chat:

```
nancy.swope@icloud.com   is_read=0 inbound rows            : 1
                         rows newer than the chat cursor   : 0
imsg-read unread_on_mac(chat)                              : 0
```

So the conversation is unread on the phone, unread by `is_read`, and invisible
to `imsg-read` in both `--chat` and `--all`. The user-visible version is "I
marked it unread on my phone and Blip never noticed, and nothing I do in Blip
puts it back".

Worth noting your fix needs both halves you wrote, not just the predicate: the
Mac-side count alone would not have made Blip show it unread, because the local
`readMark` is still newer. That is the part of this PR I had not appreciated
until I hit it.

## The one caveat on my evidence

I cannot test the other direction. The predicate this replaces exists because
chat.db carries ghost `is_read=0` rows years old, and my gateway Mac was
rebuilt on 2026-09-15 — chat.db is four days old and the history came back
from iCloud clean:

```
inbound is_read=0 rows, all time : 18
  newer than the chat cursor     : 16
  OLDER than the cursor          :  1   <- the row I just marked, today
```

One below-cursor row and it is the deliberate one. That is not evidence the
trailing-run rule is safe; it is evidence that my machine cannot show you the
case where it matters. Someone with a long-lived chat.db should check that
before this lands.

## A suggestion, and it is only a suggestion

`read_state.py` and the count change are ~60 lines of this diff and they fix a
confirmed bug with a reproduction. The PR around them is +2202/-345 across 32
files, and also carries the durable queue, a second worker process, four icons,
`scripts/demo/`, and a change to what counts as unread for reactions.

Would you consider splitting the read-state half out on its own? Selfishly I
would like the bug above fixed sooner than the whole design can be reviewed.
There is precedent from this repo two days ago — #79 was split into #90 and
#91 at @nixfred's request, and the contact-save half landed while the other
half got a longer conversation.

@nixfred, entirely your call whether that is worth the extra round.

Separately, and sorry to be the one mentioning it: #106 landed yesterday and
rewrote `imsg-read`, `pushReadArgs`, the README's `push_read` section and the
same CLAUDE.md bullet this PR replaces, and #108 changed `unread_on_mac` on top
of that. Your rebase got harder overnight through no fault of yours, and I
would rather you heard it here than found it in a merge conflict.

nixfred:
Checking in, @damonjanis. This still needs a rebase onto main before I can review the rest. Since the 18th, main has also taken #109 and #111 and shipped 2.6.0 and 2.6.1, so expect `CHANGELOG.md` to conflict at the very least.

The badge answer from the 18th still stands: count unread conversations. Ian's comment above, if you haven't seen it, is live evidence for one of this PR's own premises. An explicit unread set from the iPhone can sit below `last_read_message_timestamp`, and on `main` today that is a silent dead end.

If you've run out of time for it, say so and I'll take the rebase over rather than let it rot. Credit stays yours.

PR #104, opened 2026-09-18
docs: one Linux-side shim line in the shape diagram, listing all four shims

ianswope:
#100 left two `(Linux side)  bridge/linux/blip-shim` entries stacked in CLAUDE.md's shape diagram. The first one — the one a reader hits first — lists three shims, `{imsg,imsg-send,contacts}`, with no `imsg-read`.

`scripts/blip-setup` installs four:

```sh
for t in imsg imsg-send imsg-read contacts; do
```

and says so in its own header comment and success message.

This collapses the two entries into one: the four-shim list, the `bin_dir=` note from #100, and the bridge.conf/`shimPath()` lines that already followed.

Docs only, no code touched.


nixfred:
Merged as 6125a2b. Thanks @ianswope — that was our own mess from the bin_dir merge (313326e), two versions of the same line, one of them already out of date. The surviving line lists all four shims, which is what scripts/blip-setup actually installs. bun test 584 pass.

PR #105, opened 2026-09-18
Route per-conversation spawns through a source-id seam (#70)

### LINKED ISSUE #70: Would you take a source-routing seam, so a second messenger can plug in?
Blip's Linux side already speaks one clean dialect: every module shells out to `~/bin/imsg <cmd> --json` and parses the result into `ImsgMessage` / `ChatInfo`. That contract turns out to be almost source-agnostic already — `ImsgMessage.id` is documented as "supplied by bridges that expose it", and `service` is a free string.

I have a fork reading a second messenger (WhatsApp, through a local HTTP bridge) by writing a tool that answers that same argv and printing the same rows. Everything above the bridge — `buildThreads`, the unread ledger, toasts, `decorate`, search, the window — works unchanged.

**The question is only about the seam, not about my second messenger.** That stays in my fork; I am not proposing WhatsApp code here.

The seam is one new file and a handful of one-line call sites:

- `source-id.ts` — pure, no dependencies, built to `SourceId.mjs` the way `send-state.ts` builds `SendState.mjs`, so QML and the TypeScript share one definition. It holds `sourceFor(chat)` and `bridgeArgv(chat, tool, home, script)`, which returns the argv prefix a caller puts in front of its own. For iMessage that prefix is `["$HOME/bin/imsg"]`, `["$HOME/bin/imsg-send"]` or `["$HOME/bin/imsg-read"]` — byte-identical to today.
- Each spawn site becomes `runner(bridge.cmd, [...bridge.args, ...args], opts)` instead of `runner(\`\${HOME}/bin/imsg\`, args, opts)`: `collector.ts` (recent / chats / groups / pushRead), `thread.ts`, `avatar.ts`, `fetch.ts`, `search.ts`, and the send in `BlipView.qml`.

With no second source configured, `sourceFor` always returns `imessage` and every one of those resolves to exactly the path it resolves to now, so it is a no-op for anyone running stock Blip.

Four things in the current code assume one source, and would need widening for any second one (all four are small, and all four are arguably worth doing on their own merits):

1. `isGroupChat` treats any id containing `@` as a DM, so a non-email group id reads as a person.
2. `normalizeSendService` falls through to `"iMessage"` for anything it does not recognise, rather than saying it does not know.
3. `fetch.ts` gates attachment ids to `/^[0-9]{1,18}$/`, a chat.db rowid.
4. `BarWidget.qml` has one `watch` Process and one `watchAlive` flag driving the 6s → 60s poll stretch.

Happy to send this as a PR against `main` if it is something you would take — or to keep it in my fork and rebase forever, which is genuinely fine too. Asking first because it touches ten files by a line each, and a PR nobody wants is a PR nobody should have to read.

dreinecke:
Closes #70 — the seam only, on the terms in that thread. The second messenger stays in my fork.

**What it does.** Every spawn that is about one conversation asks `source-id.ts` which bridge answers for that id: `loadThread`, `fetchChatRows`, `pushRead --chat`, `fetchAvatar`, `sendFile`, and the text send in `BlipView.qml`. TypeScript goes through a new `bridgeFor()` beside `shimPath()` in `shim-path.ts`; QML through `SourceId.bridgeArgv()`, built from `source-id.ts` the way `BinDir.mjs` is built from `bin-dir.ts`. A call site keeps passing exactly the arguments it passed before and puts the answer in front of them.

**The iMessage path is unchanged.** Stock Blip registers no extra source (`EXTRA_SOURCES` is empty), so `sourceFor()` returns `IMESSAGE` for every id and `bridgeArgv()` is `[shimPath(tool)]` — the same program in `bin_dir`, the same argv.

**Tests (`source-id.test.ts`)**
- every Mac id shape (phone, email, group GUID, `chat<rowid>`, short code) and a set of shapes the Mac never produces, times all three tools, resolve to exactly the shim in `bin_dir`;
- `bridgeFor()` equals `shimPath()` with nothing in front, `bin_dir=` included;
- `loadThread` and `fetchChatRows` hand the runner the same command and argv as before;
- an id from a source Blip does not know routes to that source while every Mac id still reaches the Mac, and its rows are keyed, counted and threaded untouched by `chatKey` / `unreadCounts` / `buildThreads`;
- `SourceId.mjs` agrees with `source-id.ts` (also checked by loading it in Qt 6's QML engine).

`bun test`: 592 pass, 0 fail.

**Deliberately not here**
- Spawns that name no conversation — `recent`, `chats`, `groups`, `search`, `watch`, `contacts` — stay on `shimPath()`. Merging a second source's lists is the fork's business, not routing.
- The attachment fetch stays on `shimPath()`: its id is a chat.db rowid, not a conversation, and `fetch.ts` still gates on that shape.
- The four widenings from #70 (`isGroupChat`'s `@` test, `normalizeSendService`'s fallthrough, the rowid gates in `fetch.ts`/`thread.ts`, the single `watchAlive`). Each would change what the iMessage path does with an input it does not recognise, so none belongs in a byte-identical PR. The round-trip test says so where it skips asserting a DM thread's `service`.

Rebased on `c4276bf`. `CHANGELOG.md` has an Unreleased line, and `CLAUDE.md` names the seam beside the `shimPath()` invariant; the "one source for release" note next to the seam is yours to word.

https://claude.ai/code/session_01UcwAYkdiU6DaGjdUMbGYWK


nixfred:
Merged as ebefe1a. Thank you @dreinecke — framing this as the seam alone, with your second messenger staying in your fork, is what made it an easy yes, and the PR delivered exactly that: no WhatsApp code, no second source, EXTRA_SOURCES empty.

Two things I checked rather than assumed. `SourceId.mjs` is byte-identical to a fresh `bun build` of `source-id.ts`, so the QML module is genuinely generated and not a hand-edited copy that can drift. And I ran the collector against the real bridge on your branch: same unread count, same thread count, no error, which is the claim that matters most here, that the stock path is untouched.

bun test 592 pass, including your eight source-id cases. Only conflict was CHANGELOG.md against #106, which landed minutes before you; both entries kept.

The comment at the top of source-id.ts saying it is not a plugin system and carries no guarantees beyond "stock Blip resolves exactly as it did before" is the part I would have had to write myself if you had not. Thank you for that.

PR #106, opened 2026-09-18
Push group read status to Messages and restore Mac focus

jondkinney:
## What

With `push_read=thread`, reading an unread group in Blip now pushes that read to Messages on the Mac, just like a DM. The bridge selects groups through `imessage:open?groupid=...`, selects DMs through the escaped `address` parameter, and restores the previously focused app after success or failure.

This contains only read-sync work. Multi-recipient composition and group creation are separate work and are not included.

## Why

The collector previously rejected group identifiers, and the Mac helper assumed groups could not be selected by URL. This uses their opaque chat identifiers, serializes Messages selection/menu actions with an owner-only lock, and refuses to click when URL dispatch fails or Messages has no accessible window.

The default policy remains `all`; per-conversation pushes require `push_read=thread` and the updated Mac `imsg-read` tool. Local read marks remain independent, pushes still happen on unread-to-read transitions, and the bridge never substitutes a group's last speaker or writes to `chat.db`.

## How it was verified

- 584 Bun tests pass, with the machine's personal spelling dictionary isolated from test fixtures.
- Nine synthetic Mac read-routing, URL-escaping, lock, and focus-restoration tests pass.
- All nine existing read-verification tests from #102 pass; its requirement for verified completion is preserved.
- Python compilation and `git diff --check` pass.
- Read sync was confirmed working by the user on the local build before this rebase. No additional live Mac reads or sends were performed for publication; group-specific routing is covered by synthetic tests.
- The QML change only corrects a stale comment; there is no visual change requiring a screenshot.

Rebased onto current `main` (`c4276bf`). Related to #103: that PR adds durable reconciliation and conversation actions while retaining DM-only per-thread Mac pushes. This PR adds group addressing and focus restoration; the collector/helper overlap will need to be reconciled when both are integrated.


ianswope:
The group addressing and the focus restore both read right to me, and putting
`restore_front` in a `finally` is the detail I would have got wrong. One thing
I think is a bug rather than a preference, and it sits in the part that decides
whether the push worked.

`--chat` verifies through `unread_on_mac(argv[1])`, which filters
`AND c.chat_identifier = ?` (`bridge/mac/imsg-read:75` on main — this PR does
not touch the query or `UNREAD_SQL`). That is an exact match on a single `chat`
row, and the identifiers this PR now accepts are not always one row:

- A re-keyed group keeps its old `chat` rows and shares one `group_id`. That is
  what `_group_cluster_map` (`bridge/mac/imsg:1519`) and `pin_order_for` exist for.
- Messages also merges a phone SMS row with an email iMessage row under one
  `group_id` while keeping two distinct `chat_identifier`s.

So for a conversation with alias rows, when the unread landed on an alias
rather than on the identifier Blip pushed:

1. `before` is 0,
2. the `--chat` branch prints `nothing unread` and returns,
3. the push never happens — and it exits 0.

A silent no-op, reported as success. The same scoping applies to
`settle(before, argv[1])` afterwards: a cluster that Messages clears only
partly still settles to 0 for the single row being watched, and reports
`marked (0 unread)`.

This is not new to your PR; `--chat` has always had it. But it was bounded
before, because only DM handles could be addressed and the merged phone/email
DM was the one shape that could bite. Accepting group identifiers widens it to
exactly the conversations most likely to have been re-keyed.

`imsg` already resolves clusters, so the fix is presumably to expand the
identifier to its cluster's `chat.ROWID`s and count across all of them, rather
than matching `chat_identifier` once. `imsg-read` is standalone today (no
import from `imsg`), so that is a decision about where the cluster logic should
live as much as a patch.

Worth flagging that the tests cannot catch this as written: `test_read.py`
patches `unread_on_mac` wholesale, so the count is synthetic in every path. A
fixture chat.db with two `chat` rows sharing a `group_id` would pin it.

nixfred:
Merged as afedad8. Thank you @jondkinney — this one overturns a note we had been repeating as settled. Our invariant said a group "cannot be addressed at all" because `imessage://` takes a handle and a group id has no URL form. True of that scheme, wrong about Messages.

I verified the load-bearing claim on the gateway Mac (macOS 27.0) rather than taking it on trust: recorded the frontmost app, ran `open "imessage:open?groupid=<32hex>"` for a real group, asked System Events for Messages' front window title, and got the group's participant list back, then restored the front app. `open` exited 0. The deep link does exactly what you said.

The rest as merged: bun test 584 pass, 10 Mac bridge suites OK including your test_read.py (fully mocked, never opens Messages) and all nine read-verification tests from #102, whose verified-zero-or-fail contract survives. The owner-only flock and the "no accessible window" refusal are the details that make this safe to run unattended.

What I did not test: an end-to-end group push, because that marks a real conversation read on a live account and phone. Your own confirmation before the rebase stands for that.

One to watch: #103 touches the same helper and collector with DM-only pushes plus durable reconciliation. Whoever lands next reconciles by hand; I have noted it in the merge.

ianswope:
Flagging that this merged a few minutes after the comment above, so it very
likely went by unseen rather than considered and dismissed — no complaint,
just that the finding is still live on `main` at afedad8:
`bridge/mac/imsg-read:115` is unchanged.

Filed as #107 and fixed in #108, since it is a bug on `main` now rather than a
review note on a closed PR. The fix scopes the count to the conversation
cluster through `imsg`'s existing lookup, promoted to a public
`chat_cluster_ids()`, plus the fixture chat.db the old tests could not express.
`bun test` 584 pass, all 13 bridge suites OK.

Nothing here argues against what landed — the deep link and the focus restore
are the right shape, and the group path is the reason this is worth fixing
now rather than later.

PR #108, opened 2026-09-18
fix: a read push counts the whole conversation, not one chat row

### LINKED ISSUE #107: A read push verifies one chat row, so an alias's unread makes --chat exit 0 without pushing
`imsg-read --chat` verifies itself against a single chat row, so a read push
aimed at a conversation Messages has split across rows can do nothing and
still exit 0.

## Where

`bridge/mac/imsg-read:115` (on `main` at afedad8):

```python
    if handle:
        sql += " AND c.chat_identifier = ?"
        args = (handle,)
```

An exact match on one `chat` row. The identifiers Blip pushes are not always
one row:

- A re-keyed group keeps its retired `chat` rows — that is what
  `_group_cluster_map` (`bridge/mac/imsg:1519`) and `pin_order_for` exist for.
- Messages merges a phone SMS row with an email iMessage row under one
  `group_id` while keeping two distinct `chat_identifier`s.

## What happens

When the unread landed on an alias rather than on the identifier Blip pushed:

1. `before = unread_on_mac(argv[1])` is 0,
2. the `--chat` branch prints `nothing unread` and returns,
3. the push never happens — and it exits **0**.

A silent no-op recorded as success, including in `push-read.log`. The same
scoping applies to `settle(before, argv[1])` afterwards: a cluster Messages
clears only partly still settles to 0 for the single row being watched and
reports `marked (0 unread)`.

This predates #106 — `--chat` has always had it — but it was bounded while
only DM handles could be addressed, and the merged phone/email DM was the one
shape that could bite. Now that group identifiers are addressable it covers
exactly the conversations most likely to have been re-keyed. It presents as a
push that never arrives, with nothing in the log to say so, which is the
failure mode the count exists to rule out.

## Why no test caught it

`bridge/mac/test_read.py` patches `unread_on_mac` wholesale, so the count is
synthetic in every path. Nothing exercises the query against a chat.db where a
conversation has more than one row.

## Fix

`imsg` already resolves clusters for the sidebar, for pins and for
`thread --chat`; `_group_photo_chat_ids` is the lookup this needs. Promoting
it to a supported entry point and scoping the count to the cluster is a small
change; keeping a second copy of the rules in `imsg-read` would let them drift
from the bridge's own, which is the objection raised on #91.

Patch and fixture-backed regressions in the PR below.

ianswope:
Closes #107.

`imsg-read --chat` verified itself with `AND c.chat_identifier = ?` — an exact
match on a single `chat` row. Messages splits one conversation across several:
a re-keyed group keeps its retired row, and a merged 1:1 keeps a phone row
beside an email row, each with its own `chat_identifier`. Blip pushes one of
them.

When the unread sat on an alias, `before` was 0, the `--chat` branch printed
`nothing unread` and returned, and the push never happened — exit 0, no error,
nothing in `push-read.log` to suggest otherwise. `settle()` had the same
scope, so a cluster Messages cleared only partly still settled to 0 for the
one row being watched and reported `marked (0 unread)`.

The count is the referee for the whole push, so it has to cover every row the
push can clear.

## The change

- **`bridge/mac/imsg`**: `_group_photo_chat_ids` becomes `chat_cluster_ids()`,
  a supported entry point, with the private name kept as an alias so the
  existing `cmd_avatar` call site is untouched.
- **`bridge/mac/imsg-read`**: loads the vendored `imsg` beside it once, and
  scopes the count to the cluster. Falls back to the pushed identifier alone
  when the cluster cannot be resolved, which is no worse than not asking.

I kept one implementation rather than copying the rules into `imsg-read`: a
copy would drift from the bridge's own clustering, and that is the objection
you raised on #91's `phone_matcher()`. A `sync-bridge.sh` pull that renames
`chat_cluster_ids` now breaks `imsg-read`'s import loudly at load time instead
of quietly narrowing every push back to one row. Recorded in
`bridge/BRIDGE-VERSION` as a deviation, and it is a small one to upstream.

The import costs ~40 ms and happens only on the `--chat` path.

## Fail-then-pass

`test_read.py` patched `unread_on_mac` wholesale, so no test could see this.
The new cases run against a fixture chat.db. Pinning `cluster_ids` back to the
single pushed identifier, everything else unchanged:

```
===== BEFORE =====                         ===== AFTER =====
FAIL: test_merged_dm_counts_the_            Ran 13 tests — OK
      unread_on_its_other_handle
FAIL: test_rekeyed_group_counts_the_
      unread_on_its_retired_row
Ran 13 tests — FAILED (failures=2)
```

Four cases: a merged phone/email DM, a re-keyed group, two unrelated chats
that must **not** fold together (the widened scope must not quietly become a
mark-all), and an unreadable cluster falling back to the pushed row.

## Gates

| Gate | Result |
|---|---|
| `bun test` | 617 pass, 0 fail, 20 files |
| all 14 `bridge/mac` suites in `ci.yml` | OK |
| `test_group_photo.py`, `test_cluster.py` (the renamed helper's callers) | OK |
| `python3 -m py_compile bridge/mac/imsg bridge/mac/imsg-read` | OK |

## Not claimed

No live Mac in this loop: the fixture is synthetic and the UI calls are not
exercised. What this proves is the count, which is the part that was wrong.
I have the gateway Mac and can run a real round trip on a merged DM before you
merge if you would rather have that first — say the word.

---

Rebased onto `de29ccb` (#91, #105 and 4134bbf landed while this was being written). Only `CHANGELOG.md` conflicted, additively; gates above are from after the rebase.


nixfred:
Merged as 1a4de0a. Thank you @ianswope — this is the kind of find that matters most: #102 made the Mac's own count the referee for a read push, and a referee watching one row of a cluster is worse than no referee, because it reports a clean exit for a push that did nothing.

It is also our own invariant that the verification was not following. CLAUDE.md already says pins and merged 1:1s use the whole conversation cluster, and that a read of a canonical conversation covers its aliases. The push just never got the memo.

Verified here: bun test 617 pass, 11 Mac bridge suites OK, chat_cluster_ids loads as a supported entry point with the private alias preserved so cmd_avatar is untouched. What I could NOT show live: the gateway Mac has no chat_identifier with more than one chat row today, so the cluster case does not reproduce on it. Your synthetic tests carry that.

Deployed to the gateway Mac along with #106's group addressing. Only conflict was CHANGELOG.md against the four merges earlier today.

ianswope:
Closing: this is on `main` as 1a4de0a. The open state and the conflict badge are
artefacts of the merge landing by hand rather than through the button — the
branch here carries a rebase of the same commit, and `main`'s copy is identical.

Verified on `main`: `chat_cluster_ids()` in `imsg`, `cluster_ids()` and the
cluster-scoped `IN` clause in `imsg-read`, and all 13 `test_read.py` cases.

PR #110, opened 2026-09-19
fix: autofill types the code when Gecko ignores set_text_contents

joshhattan:
## What

Security-code autofill now works in Zen and Firefox. After the accessibility text write, the helper re-reads the field; if it is still empty it types the code through the Hyprland key-event path Chromium already uses. A write that did land keeps the existing early return, so digits are never entered twice. Adds a CHANGELOG entry and updates the Zen note in `docs/AUTOFILL.md`.

## Why

In Zen, clicking **Fill code** did nothing and showed no error. Gecko's AT-SPI `set_text_contents()` returns `True` but leaves the field empty, and `fill()` returned straight after the call without checking. Everything before that point was correct: `field()`/`snapshot()` returned the right metadata, `classify()` gave `smart`, the offer was emitted and the accept token was valid.

Confirmed on a throwaway page whose `oninput` handler writes the value and an event count into `document.title`:

| step | field | title |
|---|---|---|
| `set_text_contents("482913")` returns `True` | empty | unchanged (no `input` event) |
| `fill()` with this change | `482913` | `got=482913 inputEvents=6` |

The key-event path keeps all its per-character guards (focus, `EDITABLE`/`SHOWING`/`SENSITIVE`, window address and pid, session lock, deadline).

## How it was verified

- [x] `bun test` is green (621 pass on current `main`), and `python3 test_otp_desktop.py` passes (8 tests)
- [x] Any send/receive behavior was exercised against **my own number only** — not applicable: no send/receive code touched; the codes in the tests and page are invented
- [ ] UI change → screenshot attached — not applicable: no UI change
- [x] Touches an invariant in `CLAUDE.md` → none. No message content is stored, and code text still never reaches argv or QML.

New tests (`SingleField` in `test_otp_desktop.py`): an accepted write does not also type, and an ignored write falls back to typing. The second fails without the change.

Also verified end to end on a real sign-in page in Zen 1.22.2b.

Environment: Blip 2.5.0 · Omarchy 4.0.0.r2158.gd174d4a-1 · quickshell 0.3.1 · Hyprland 0.56.2 · Zen 1.22.2b · macOS 27.0


nixfred:
Merged — thank you. This is the kind of bug that is hard to even see: the write reported success, so every log line upstream of it looked right, and the only symptom was a button that did nothing. Reading the field back is the right referee, and keeping the early return when the write did land is what stops the double entry.

Verified here before merging: 621 `bun test` and 8 `python3 test_otp_desktop.py` green, and your `test_ignored_write_falls_back_to_typing` fails on `main` without the change, which is exactly what a regression test should do. Neither Zen nor Firefox is the browser on this machine, so the live half rests on your sign-in page and your `input`-event count — recorded as such in the weekly notes rather than claimed as ours.

One thing left open, not a blocker: Gecko's `set_text_contents()` is synchronous, so a write that landed late would arrive alongside the typed keys and double the code. Unlikely, and nothing re-checks between the two. If you want to harden it, a single re-read before the first key event would close it.

Also: we pair every Blip contributor with an X account in `CHANGE.log`, and your GitHub profile publishes nothing we can verify. If you would like to be credited with a handle, drop it here and it goes in the weekly post. "Rather not be tagged" is an equally good answer — you are logged as "none found" until you say otherwise.

PR #111, opened 2026-09-20
docs: CLAUDE.md deploy recipe and shim list match the code

ianswope:
Four stale facts in `CLAUDE.md`. Two of them silently cost working features on a machine deployed by following it — both found while bringing a box from 2.5.0 up to 2.6.1 this morning.

### The deploy `cp` omitted `otp-desktop.py`

`otp-autofill.ts:22` spawns the helper from `import.meta.dir` — the installed plugin directory. A deploy made from this recipe therefore leaves security-code autofill with no helper to spawn, and nothing says so. `AGENTS.md:43` and `CONTRIBUTING.md:42` already list the file; only the `CLAUDE.md` copy was missing it, so the three recipes now agree.

### The Linux shim list stopped at four names

`blip-setup` has installed `contact-save` as a fifth shim since 39bcd29 (#91). While the shape diagram still said four, the Mac could end up holding `contact-save` in `~/.blip/bin` with nothing on the Linux side able to reach it — which is exactly what I found here: the Mac tool present and current, `~/bin/contact-save` absent.

### Two counts

- `blip-dispatch` gates six tools, not five — `TOOLS` at `bridge/mac/blip-dispatch:21`.
- `bun test` is 623 tests in ~1.5 s, not "90+ tests, ~40 ms".

Docs only, no behaviour change. `bun test` green at 623 pass / 0 fail.


nixfred:
Merged, thank you. This turned out to be more than a docs fix for us.

I had used exactly this recipe to take vic from 2.5.0 to 2.6.1, so its `otp-desktop.py` was still the 2.5.0 copy and #110's Gecko fallback never reached it. Neither of our two machines had `~/bin/contact-save` either, so Save-as-contact from #91 failed on both. Checked after merging: on both hosts `contact-save` now reaches the Mac and gets the tool's own `{"ok":false,"code":"invalid"}` for empty input, where it used to be refused locally. All five shims were also four commits stale (Sep 1), so they've been refreshed. `imsg recent` and `imsg chats` exit 0 through the new shim.

Every claim checked against the code first: the `TOOLS` set, the `import.meta.dir` spawn, the `blip-setup` shim loop, and the other two recipes.

